PR #28, opened 2015-01-29
load_value returns unicode instead str

marc1n:
I've changed load_value function ito return unicode instead of str. It is important when we use non ASCII characters.

Now we have:
>>> print toml.loads('key = "val"')
{'x': u'a'}

Before my change we had:
>>> print toml.loads('key = "val"')
{'x': 'a'}

### DIFF
--- a/toml.py
+++ b/toml.py
@@ -50,7 +50,7 @@ def loads(s):
         keyname = False
         delnum = 1
         for i in range(len(sl)):
-            if keyname and not openstring:                
+            if keyname and not openstring:
                 if sl[i] == '\n':
                     raise Exception("Key name found without value. Reached end of line.")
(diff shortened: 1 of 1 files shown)


uiri:
This change breaks python3 compatibility.
This change fails the unicode-escape test:

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\u03b4' in position 0: ordinal not in range(128)
```

I would like to return unicode instead of str in python2, but it is difficult to do so while maintaining python3 compatibility.

uiri:
Seeing as there have been no updates on this in over a year, I am closing this PR.
If you believe this is in error, feel free to re-open the PR or to submit a new PR.

PR #29, opened 2015-01-30
Remove None values when dumping Mapping

jsancio:
dictionary can sometimes contain keys with a None value. This leads to the generation of a TOML string with None in it. This is not valid TOML.


uiri:
This looks good to me.
Let me know if you want a copyright line in the LICENSE file.

PR #32, opened 2015-02-05
Fix endless loop

octycs:
The following toml file would cause an endless loop (somevalue is not in quotes):
```toml
somekey=somevalue
```
Because more than one '=' is not specified in [spec](https://github.com/toml-lang/toml),
`line.split()` should only be called with *count=1*.
--> You don't need to increase *count* infinitely if data type is unknown.


uiri:
As of v0.3.1, I believe that = is allowed inside key names. Your change forbids = in keynames. That was the reason for skipping equal signs until the last one is found in the first place.

octycs:
Sorry. It's fixed now, so I'll close this.

PR #44, opened 2016-05-05
Recognize both str and bytes as string in Python 3

bkabrda:
With this pull request, all the places that only accept `str` now will also accept `bytes`, which I believe is the correct thing to do.

### DIFF
--- a/toml.py
+++ b/toml.py
@@ -20,7 +20,7 @@ try:
 except NameError:
     unicode = str
     _range = range
-    basestring = str
+    basestring = (str, bytes)
     unichr = chr
 
 def load(f, _dict=dict):


uiri:
I'm not sure that this is the right thing to do. Per [PEP 358](https://www.python.org/dev/peps/pep-0358/), Python2 strings are overloaded to hold both characters and bytes. In Python3, strings are sequences of characters only and there is a separate type for sequences of bytes. Since a TOML file must contain only UTF-8 encoded characters, I think accepting a sequence of bytes may not be the right thing to do.

I am going to think about this some more but the fix as it stands seems kind of ugly and I'm not sure that it won't have unintended side effects.

bkabrda:
IIUC accepting sequence of bytes is correct, assuming you try to decode them with utf-8 [1] (and if that fails, then the input is incorrect).

[1] https://github.com/bkabrda/toml/blob/master/toml.py#L51

uiri:
> IIUC accepting sequence of bytes is correct, assuming you try to decode them with utf-8 [1] (and if that fails, then the input is incorrect).
> 
> [1] https://github.com/bkabrda/toml/blob/master/toml.py#L51
> 

I have been on the fence after thinking about this for a while. Since the initial API is essentially the same as that of python's builtin JSON module, I looked at what that module does. That module doesn't seem to like bytes either. For that reason, I recommend you try the same workaround in your downstream code and report a bug if that doesn't work :)

Sorry, I'm closing this as WONTFIX.

PR #45, opened 2016-05-06
add travis testing

nikolas:


.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: python
+python:
+  - "2.7"
+  - "3.5"
+install:
+ - python setup.py install
+script:
+ - cat examples/example-v0.4.0.toml | python toml_test.py
```
uiri: Any reason why this isn't `python toml_test.py < examples/example-v0.4.0.toml` ?

nikolas: Not really.. that way makes more sense and also works! I'll change it now.

PR #48, opened 2016-05-17
Peephole optimizations.

ProgVal:


uiri:
This is generally a library which other people will import so I try not to make any global changes. For that reason I would prefer that anything which is global within the file that is not a part of the public API be prefixed with an underscore per PEP 8. The public API consists of `load`, `loads`, `dump` and `dumps`. I suppose I should be more consistent about this by renaming my helper methods.

isnumeric() and isigit() return true for unicode characters outside of the ASCII 0-9 range so using them does not preserve previous behaviour. The backslash code you've written is difficult for me to verify with the interpreter in my head. Obviously, I find the way I wrote it to be much simpler and YMMV. I'm hesitant to use such a complicated backslash regex.

There are a couple of minor typos around line 300 where it should be: `candidate = candidate_group + ", " + candidate_groups[0]`
I realize that it isn't the easiest code to read but it is hard to do a for...in... loop while also modifying the list that is being looped over.

Apologies for the delay in reviewing this.

ProgVal:
On 05/07/2016 04:02, W Pearson wrote:
> I'm hesitant to use such a complicated backslash regex.

It's the same regexp, except I combined all characters in a single class
instead of having one regexp per character.

ProgVal:
On 05/07/2016 04:02, W Pearson wrote:
> There are a couple of minor typos around line 300 where it should be: `candidate = candidate_group + ", " + candidate_groups[0]`

Are you sure? From what I understand, the previous code removed the
first item, added `candidate_group + ','` before, and pushed it back in
front of the list.
My new code directly prepends `candidate_group[1:] + ','` (which is
exactly `candidate_group + ','` in the old code, as I did not pop the
first item) to the first item.

ProgVal:
On 05/07/2016 04:02, W Pearson wrote:
> isnumeric() and isigit() return true for unicode characters outside of the ASCII 0-9 range so using them does not preserve previous behaviour.

Actually, that's only for isnumeric, according to the doc.

But, looking at the code, I noticed I replaced `re.match('^[0-9]',
value)` with `value.isnumeric()`, which is indeed incorrect.
Fixed.

uiri:
The implementation (ie: non-public api) global variable names still need underscore prefixes.

The documentation for isdigit which I see for python 2 says that it is locale-dependent for 8-bit strings.
The documentation for isdigit which I see for python 3 says that other characters, e.g. superscript digits are also included.

I think having one regexp per character is easier to read and understand.

candidate_groups is the line split on commas which is a list. candidate_group is the current group being parsed/built up which is a string. I think the re-written code should be something like this:

candidate_groups[0] = candidate_group + "," + candidate_groups[0]

ProgVal:
On 09/07/2016 05:51, W Pearson wrote:
> The implementation (ie: non-public api) global variable names still need underscore prefixes.

Done.

> The documentation for isdigit which I see for python 2 says that it is locale-dependent for 8-bit strings.
> The documentation for isdigit which I see for python 3 says that other characters, e.g. superscript digits are also included.

Indeed. Fixed.

> I think having one regexp per character is easier to read and understand.

The difficulty of the regexp comes from the beginning. I don't think having `+escapes[i]` or `[%s]" % ''.join(escapes)` affects the readability much.
But it does improve the performance.

> candidate_groups is the line split on commas which is a list. candidate_group is the current group being parsed/built up which is a string. I think the re-written code should be something like this:
> 
> candidate_groups[0] = candidate_group + "," + candidate_groups[0]

Indeed. I was confused by the two variable names being very similar.

uiri:
> > The implementation (ie: non-public api) global variable names still need underscore prefixes.
> 
> Done.

Your original commit contained one instance of this which you missed. I amended that commit, rebased the rest of this PR on top of master and the amended commit.

Closing.

PR #51, opened 2016-05-18
Do not raise base Exception.

### LINKED ISSUE #50: Base Exception are used, instead of subclasses

ProgVal:
Closes #50.

Note: you should merge #48 first, as this PR is based on it.

### DIFF
--- a/toml.py
+++ b/toml.py
@@ -3,6 +3,9 @@
 import re
 import datetime
 
+class TomlDecodeError(Exception):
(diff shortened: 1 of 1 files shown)


uiri:
> Note: you should merge #48 first, as this PR is based on it.

Is there any reason that this is the case? This seems a lot more straightforward than #48.

ProgVal:
as you wish

PR #57, opened 2016-08-08
add types

tharvik:
Add a stub file to ease type checking.

As defined in [PEP-484](https://www.python.org/dev/peps/pep-0484/), python now support type checking. It can be specified either via function annotation or stub files. To keep compatibility with py2, it is a stub.

It is useful for development and for other devs to allow typechecking their code using toml.

It can be tested with `mypy .` and that will test that effectively, `toml.py` is compliant to `toml.pyi` and code using it (as it also check other files there, it currently fail on `toml_test.py` which can be fix by having a check on `sys.version_info < (3,)` instead of a check on `NameError`).

There isn't currently an [official way](https://github.com/python/mypy/pull/1895) to install these stubs to be used by mypy, but it will soon be there.


uiri:
> (as it also check other files there, it currently fail on `toml_test.py` which can be fix by having a check on `sys.version_info < (3,)` instead of a check on `NameError`).

Do you want to make this change before I merge this in?

tharvik:
> > (as it also check other files there, it currently fail on `toml_test.py` which can be fix by having a check on `sys.version_info < (3,)` instead of a check on `NameError`).

> Do you want to make this change before I merge this in?

Done.

uiri:
Merged.

Issue #64, opened 2016-09-22
Error parsing strings with curly brackets


matthew-brett:
For this file:

```
command = ['{text1}', "text2"]
```

I get

```
In [17]: toml.load(open('my.toml', 'rt'))
Out[17]: {'command': ['{text1']}
```

The list has been truncated at the first closing curly bracket.

takluyver:
I have made #66 to fix this specific bug, but having looked at the code to do so, I don't have much confidence that there aren't similar bugs remaining. It uses a lot of string munging and special casing, and the tests that are configured to run on Travis don't test much. I'm going to look at using pytoml instead.

uiri:
There is a test suite - https://github.com/BurntSushi/toml-test is the original and a fork with a number of additional tests is maintained by avakar https://github.com/avakar/toml-test

Neither test suite is as well maintained as I would like so I rely primarily on bug reports to find issues.
The other problem is that often someone will send me a pull request which appears to fix an issue but doesn't cover certain corner cases.

Reopening due to issue uncovered by a test in BurntSushi's test suite which isn't present in avakar's test suite: arrays of inline tables are now broken.

uiri:
(Whoops, it appears that that is an already existing issue. Sorry.)

takluyver:
There is a test suite, but your [Travis config](https://github.com/uiri/toml/blob/master/.travis.yml) doesn't appear to run it, and I don't see clear instructions on testing it (I could probably figure it out from the test repo, but it would be nice to have brief instructions in this repo).

uiri:
There is a TESTING file in the repo.
If it is unclear or you have suggested improvements, feel free to open up a new PR or Issue.

-Will

PR #66, opened 2016-11-23
Fix detection of array of objects (inline tables)

takluyver:
A string can contain '{', so `'{' in s` is not a good enough test of whether something is an inline table.

Closes gh-64


uiri:
Thanks! Merged as c45b5ff

PR #69, opened 2016-12-03
Fix #68: correctly handle whitespace in multiline strings.

### LINKED ISSUE #68: New lines in multi-line literal strings trimmed incorrectly
The library is incorrectly parsing/generating code from multi-line literal strings by condensing multiple new lines in the string into a single one.

Example:

**EXPECTED:**

```python
import toml
toml.loads("a = '''\nhello\nworld\n\n\nbye'''")
{u'a': u'hello\nworld\n\nbye'}
```

**ACTUAL:**

```python
import toml
toml.loads("a = '''\nhello\nworld\n\n\nbye'''")
{u'a': u'hello\nworld\nbye'}
```

We expect two new lines between `world` and `bye`, but we get one. According to the specification (emphasis mine):

> Multi-line literal strings are surrounded by three single quotes on each side and allow newlines. Like literal strings, there is no escaping whatsoever. A newline immediately following the opening delimiter will be trimmed. **All other content between the delimiters is interpreted as-is without modification.**

I'll take a look at the code and propose a fix.

SergioBenitez:
Previously, the library would trim all whitespace in multiline strings, even when the string was a multiline _literal_ string. This resolves that issue, leaving all whitespace in multiline literal strings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ python:
  - "3.5"
  - "3.6"
  - "pypy"
+notifications:
+  email: false
 addons:
   apt:
     packages:


uiri:
This patch looks good however, I am hesitant to accept it since it does break the toml test suite's multiline string test.

Could you submit a fix to that test to @BurntSushi upstream? After that, I will accept the PR.

SergioBenitez:
@uiri Why does it matter if it breaks the test suite if the test suite is incorrect? I've (and you've) mentioned @BurntSushi in both posts, and there are issues opened (https://github.com/BurntSushi/toml-test/issues/18) about the problem. Further, the test suite is outdated: TOML is a v0.4 while the suite tests for v0.2. If you think the changes are correct, then why does it matter if an incorrect test suite fails?

In any case, `pytoml` does the right thing here, and I'm fairly certain (though I haven't tried) that these changes will pass a fork of @BurntSushi's test suite: https://github.com/avakar/toml-test/

uiri:
> @uiri Why does it matter if it breaks the test suite if the test suite is incorrect? 

If the test suite is incorrect then it should be fixed. I rely on the test suite when I make bug fixes and add in features related to new versions of the spec. The test suite in question is also language agnostic - improving the test suite will also improve other TOML implementations. This promotes interoperability when multiple different implementations (in different languages) use the same TOML files.

> Further, the test suite is outdated: TOML is a v0.4 while the suite tests for v0.2.

I do not think that the test suite is outdated. The version which it claims to support is not necessarily the latest version which it actually supports.

> If you think the changes are correct, then why does it matter if an incorrect test suite fails?

For a number of different reasons all stemming from the fact that maintenance work on this library is sporadic and my memory is, to put it mildly, less than perfect. I'm sure I've accidentally reintroduced bugs due to missing tests for them in the test suite before. I want to prevent that from happening again.

> In any case, `pytoml` does the right thing here, and I'm fairly certain (though I haven't tried) that these changes will pass a fork of @BurntSushi's test suite: https://github.com/avakar/toml-test/

The forks of the test suite maintained by @Avakar and @BurntSushi respectively have now diverged to the point that ideally I would have *both* of them passing until @BurntSushi merges @Avakar's tests in.

uiri:
From the spec:

> For writing long strings without introducing extraneous whitespace, use a "line ending backslash". When the last non-whitespace character on a line is a `\`, it will be trimmed along with all whitespace (including newlines) up to the next non-whitespace character or closing delimiter.

Closing as WONTFIX due to compliance with the spec.

SergioBenitez:
As I mentioned in #71, that line of the spec is entirely related to the issue at hand, and as a result, to this PR.

uiri:
That line of the spec is relevant to the test which this PR breaks. Looking at the test suite, it conforms to the spec.

I am happy to accept this PR if it does not break the test suite.

uiri:
Oh wow, I'm really sorry about accidentally closing this.

A button in the GitHub Web UI merged the upstream master into your downstream master... which is something no one would ever want and I'm really not sure why the button to do that exists. In trying to fix it, I accidentally pushed the latest master to your repo's master instead of your previous commit to that master. That commit still exists in this repo in the `pr69` branch.

PR #70, opened 2017-01-02
Added minus sign to allow negative numbers in inline dicts

samvasko:
Trivial dict like this would fail to parse.

```toml
test = {x = -3}
```


uiri:
Thanks!

PR #72, opened 2017-01-19
Fixes #71: Do not strip() line if its part of multiline str

### LINKED ISSUE #71: Indentation and whitespaces are being stripped away in multiline strings
Hi,

the current implementation of the parser strips away whitespace from all lines using `strip()`.
To reproduce use this snippet:

``` python
import toml

t = '''
foo = """
test
  test
  test2
test3
    test5
    test6
test7
"""
'''

p = toml.loads(t)
print p["foo"]
```

**Acutal Output**

```
test
test
test2
test3
test5
test6
test7
```

**Expected Output**

```
test
  test
  test2
test3
    test5
    test6
test7
```

This violates the `toml` spec (see https://github.com/toml-lang/toml#string)

> Multi-line basic strings are surrounded by three quotation marks on each side and allow newlines. A newline immediately following the opening delimiter will be trimmed. All other whitespace and newline characters remain intact.

furthermore

> A newline immediately following the opening delimiter will be trimmed. All other content between the delimiters is interpreted as-is without modification.

noqqe:
Fixes #71

### DIFF
--- a/toml.py
+++ b/toml.py
@@ -190,7 +190,8 @@ def loads(s, _dict=dict):
     multilinestr = ""
     multibackslash = False
     for line in s:
-        line = line.strip()
+        if not multilinestr:
+            line = line.strip()
         if line == "":
             continue
         if multikey:


uiri:
From the spec:

> For writing long strings without introducing extraneous whitespace, use a "line ending backslash". When the last non-whitespace character on a line is a `\`, it will be trimmed along with all whitespace (including newlines) up to the next non-whitespace character or closing delimiter.

Closing as WONTFIX due to compliance with the spec.

Issue #74, opened 2017-02-02
Preserving inline tables


nateprewitt:
Hey @uiri, are there any plans to support preserving inline tables in the `dumps` function. If not would you be open to a PR implementing such functionality?

Right now, I have a project with users supplying TOML, often using the inline tables syntax:

`value = {extra = ["values",]}`

While `loads` handles this just fine, serializing back to TOML creates a subsection. This is creating maintainability problems, in some cases, since things are moving around in the TOML file unexpectedly. It would be nice if the data structure was written back out as it was ingested, if possible.

Thanks for your consideration!

uiri:
Hi Nate,

> Hey @uiri, are there any plans to support preserving inline tables in the `dumps` function. If not would you be open to a PR implementing such functionality?

I do not have any immediate plans to add this support. However, I am always open to pull requests.

> While `loads` handles this just fine, serializing back to TOML creates a subsection. This is creating maintainability problems, in some cases, since things are moving around in the TOML file unexpectedly. It would be nice if the data structure was written back out as it was ingested, if possible.

I would recommend calling `loads` with the _dict parameter set to an OrderedDict. This (or another custom dictionary type) might help your issue. If you still find the need to make changes to the toml code, I would recommend adding a boolean flag to `dumps`. That way the current behaviour can stay as the default.

nateprewitt:
> I would recommend calling `loads` with the _dict parameter set to an OrderedDict. This (or another custom dictionary type) might help your issue.

Unfortunately, we're already doing that to no avail.

> I would recommend adding a boolean flag to `dumps`. That way the current behaviour can stay as the default.

Yep, I was thinking a preservation flag for opt in, so it won't affect current users. I'll take a look into setting that up, thanks!

uiri:
Fixed at 87993ce (I believe)

PR #75, opened 2017-02-04
add preserve flag to dumps to preserve inline table syntax

nateprewitt:
This is a patch to address issue #74.

This library will currently ingest TOML in the format below with `loads`:

```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true

[dev-packages]
pytest = "*"

[packages]
codecov = "*"
pytest-httpbin = {extras = ["Flask", "fast"]}
sphinx = "==1.4.8"
```

However, when you go to save this TOML back to the file, it will rewrite the contents as:

```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true

[dev-packages]
pytest = "*"

[packages]
codecov = "*"
sphinx = "==1.4.8"

[packages.pytest-httpbin]
extras = [ "Flask", "fast",]
```

While this is valid syntax, it would be nice to preserve the more compact inline table syntax when present. This patch adds a flag `preserve` to `dumps` which will write the contents in inline table syntax back out to the file, rather than expanding it.

Some outstanding notes:
* I'd be happy to add tests for this, but didn't see a way to do it with the go test library
* I only added the flag to `dumps` and `_dump_section` but not `dump`. I could see it being useful there too but didn't want to make more changes than needed. Let me know if you'd like to see it there as well.

### DIFF
--- a/toml.py
+++ b/toml.py
@@ -711,14 +711,10 @@ def _dump_inline_table(section):
         for k, v in section.items():
             val = _dump_inline_table(v)
             val_list.append(k + " = " + val)
-        retval += "{" + ", ".join(val_list) + "}\n"
-        return retval
-    elif isinstance(section, list):
-        val_list = [_dump_inline_table(i) for i in section]
-        retval += "[" + ", ".join(val_list) + "]"
+        retval += "{ " + ", ".join(val_list) + " }\n"
         return retval
     else:
-        return "\"" + section + "\""
+        return str(_dump_value(section))
 
 def _dump_value(v):
     if isinstance(v, list):
--- a/toml.py
+++ b/toml.py
@@ -25,6 +25,10 @@ class TomlTz(datetime.tzinfo):
     def dst(self, dt):
         return datetime.timedelta(0)
 
+class InlineTableDict(dict):
+    def __init__(self, *args):
+        dict.__init__(self, args)
+
 try:
     _range = xrange
 except NameError:
@@ -520,7 +524,7 @@ def _load_value(v, strictly_valid=True):
     elif v[0] == '[':
         return (_load_array(v), "array")
     elif v[0] == '{':
-        inline_object = {}
+        inline_object = InlineTableDict()
         _load_inline_object(v, inline_object)
         return (inline_object, "inline_object")
     else:
@@ -692,7 +696,7 @@ def _dump_sections(o, sup, preserve=False):
                 if o[section] is not None:
                     retstr += (qsection + " = " +
                                str(_dump_value(o[section])) + '\n')
-        elif preserve:
+        elif preserve and isinstance(o[section], InlineTableDict):
             retstr += (section + " = " + _dump_inline_table(o[section]))
         else:
             retdict[qsection] = o[section]
--- a/toml.py
+++ b/toml.py
@@ -628,7 +628,7 @@ def dump(o, f):
     f.write(d)
     return d
 
-def dumps(o):
+def dumps(o, preserve=False):
     """Returns a string containing the toml corresponding to o, a dictionary"""
     retval = ""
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hey @uiri, checking in to see if you have any thoughts here. Thanks again for your consideration.

uiri:
Hi @nateprewitt,

The code looks fine from inspection but I'm a little concerned that the flag will create _all_ tables as inline tables. I want to do a bit of manual testing before accepting it. Unfortunately, I haven't been able to find the time over these past couple of weeks.

nateprewitt:
@uiri hmm, yeah I think you are correct. This is more of a toggle for using tables and inline tables. Is there any way to determine the format the table was ingested in currently?

uiri:
> @uiri hmm, yeah I think you are correct. This is more of a toggle for using tables and inline tables. Is there any way to determine the format the table was ingested in currently?

I do not believe that there is currently a way to determine the format the table was ingested in.

Perhaps an empty subclass of dict plus a call to isinstanceof(...) in the condition for preserving a table's contents as an inline dictionary? The subclass could conceivably override __str__ to do the right thing.

nateprewitt:
>Perhaps an empty subclass of dict plus a call to isinstanceof(...) in the condition for preserving a table's contents as an inline dictionary? The subclass could conceivably override __str__ to do the right thing.

That seems like a reasonable approach to me, I'll take a look at getting that implemented tomorrow.

nateprewitt:
Sorry this took so long to get around to. I've got the patch working now so you can have a mix of inline and standard tables with original formatting preserved.

This implementation is, at least in my opinion, the cleaner of the two options. Having the `InlineTableDict` as essentially a sentinel object is a bit silly but the `__repr__` implementation is less desirable. We'd need to rebundle all the logic for printing (`_dump_inline_table` and `_dump_value`) inside the class which creates a fair amount of redundant code. We could call the functions directly from the class but it seems wrong to use external functions for something that fundamental to the object.

We can obviously tweak this if you feel strongly about overwriting `__repr__` but this is the minimal required change to get what we want. When you've got a moment please me know your thoughts, thanks!

kennethreitz:
💫✨🍰✨💫

nateprewitt:
Hey @uiri, wanted to check in again and see if you have any further thoughts on this. Thanks!

nateprewitt:
Hey @uiri, I'm sure you're busy. Just wanted to ping here again in case this fell off the radar. Thanks again!

toml.py
```diff
@@ -25,6 +25,10 @@ def utcoffset(self, dt):
     def dst(self, dt):
         return datetime.timedelta(0)
 
+class InlineTableDict(dict):
+    def __init__(self, *args):
+        dict.__init__(self, args)
+
```
ryanhiebert: Did you mean to use `*args` in the call to `__init__`?

If I were you, I'd suggest leaving out the method declaration entirely, or at least using `super` instead of calling the `dict` implementation directly, if you intended to avoid the ordering gotcha of `**kwargs` and `OrderedDict`.

Then an analog of the `_dict` parameter on `load` and `loads` (`_InlineTableDict`?) that allows you to override which one gets used, and you can then pass in a ordered subclass, same as you do for `_dict`. For example:

```python
class OrderedInlineTableDict(InlineTableDict, OrderedDict):
    pass
```

uiri: `_dict` applies to `load` and `loads`
This class applies to `dump` and `dumps`. I don't see this as a big issue.
Feel free to submit a follow-up PR.

uiri:
Hi Nate, sorry for the delay.

Merged in at 87993ce

nateprewitt:
No worries @uiri, glad we got things wrapped up. Thanks!

Issue #78, opened 2017-02-21
Loading from list of filenames should return a list


samvasko:
Related to [this section](https://github.com/uiri/toml/blob/master/toml.py#L41-L48) of code.
When calling `.loads([])` my expectation would be to get a list, not a dictionary with all the values merged together. I do not really see a use case for the latter. It also causes unexpected rewrites. 

Example:
```Python
with open('file1.toml', 'w') as f:
    f.write('a = 1\n')

with open('file2.toml', 'w') as f:
    f.write('a = 2\n')

tom = toml.load(['file1.toml', 'file2.toml']);
print(tom) # Prints {'a': 2} 
```

RazerM:
This is used like [`ConfigParser.read`](https://docs.python.org/3.6/library/configparser.html#configparser.ConfigParser.read), usually so that a local config file can take precedence over a file in `$HOME` and/or elsewhere.

uiri:
If you want a list, why not use a list comprehension?

```python
tomlified_filelist = [toml.load(f) for f in filelist]
```

I'm hesitant to change the behaviour because although you do not see a use case for it, someone downstream may be depending upon its behaviour and I would rather not break dependency that without a good reason.

samvasko:
I guess this is more of documentation issue then 📖

uiri:
Documentation Fixed by #79 

Thanks Sam!

PR #79, opened 2017-03-01
Improve docs and public API

samvasko:
Hello, I've added comment to all *public* methods. Feel free to add your edits.

Couple more things I would like to do
- [x] ~Rename parameters of public functions to be more descriptive~
- [x] ~Make `dump` accept a filename (like `load`)~
- [x] Add the docs I just added to readme
- [x] Travis build status?

### DIFF
--- a/tests/decoding_test.sh
+++ b/tests/decoding_test.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+python tests/decoding_test.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,16 @@
 language: python
 python:
-  - "2.7"
-  - "3.5"
+ - "2.7"
+ - "3.6"
+addons:
+  apt:
+    packages:
+    - golang
+before_install:
+ - export GOPATH=~/go
+ - go get github.com/BurntSushi/toml-test
 install:
  - python setup.py install
+ - chmod +x ./tests/*.sh
 script:
- - python toml_test.py < examples/example-v0.4.0.toml
+ - ~/go/bin/toml-test ./tests/decoding_test.sh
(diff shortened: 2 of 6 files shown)


uiri:
Hi sam,

These docstrings look great. Do you want me to merge this right away or when you're done these additional tasks as well?

If we could integrate the Travis with both BurntSushi and avakar's test suite that would be good. I would be more comfortable adding some collaborators to the repo if we have automated checks for the test suites.

Thanks!

samvasko:
I was hoping someone could do a proofread 📄

samvasko:
I will look into the travis setup :)

uiri:
> I was hoping someone could do a proofread 📄 

While giving it a second look, I noticed that your description for `_dict` is incorrect. `_dict` specifies the class of the returned toml dictionary. The default is the default python `dict`.

One example use case would be to pass `toml.loads(s, _dict=OrderedDict) in order to get a consistent ordering when converting from a toml file to a dictionary and then back out to a toml file.

Otherwise, it looks good (grammar, etc.)

samvasko:
Ok I've updated the travis file. For now it only runs BurntSushi's decoding test. Hopefully it will get updated to latest spec soon.

tests/decoding_test.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+python tests/decoding_test.py
```
samvasko: This file is here because using !# does not work when using virtual environments and the toml-test does not support multiple arguments

samvasko:
I crossed out the first two tasks as I just don't feel like doing that right now. Feel free to merge this whenever. :man_shrugging:

samvasko:
closes #78

PR #82, opened 2017-03-27
fixes uiri/toml#81

zed:


uiri:
Merged in at 25e40fd

Thanks zed

PR #83, opened 2017-03-27
always read toml file using utf-8

zed:
[The spec says](https://github.com/toml-lang/toml#spec):

> A TOML file must be a valid UTF-8 encoded Unicode document.

Builtin `open()` in Python 3 uses `locale.getpreferredencoding(False)` by default. It may differ from utf-8. To read a TOML file, the utf-8 encoding should be passed explicitly.

### DIFF
--- a/toml.py
+++ b/toml.py
@@ -2,6 +2,7 @@
 
 import re
 import datetime
+import io
 
 class TomlDecodeError(Exception):
     pass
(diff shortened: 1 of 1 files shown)


uiri:
Thanks for the PR! I did a bit of testing and it looks like it works in both 2 and 3.

PR #84, opened 2017-03-27
Fix negative partial-hour offsets

ryanhiebert:
The sign of the offset needs to be applied after the timedelta is constructed, or else in the case of negative offsets the minutes will be counted in the opposite direction as the hours.


uiri:
Merged in at  309faab

PR #85, opened 2017-03-29
Refactor _dump method and replaced range(len) with enumerate

Jackevansevo:
Refactored the _dump method to lookup corresponding method based on the values current type. Prevents repeated type checks on a single variable.

Also replaced use of `_range(len(sl))` with `enumerate` to iterate over the list `sl` storing the current value in `item` and index in `i`. This prevent repeated lookups of `sl` with `sl[i]`.

Everything seems to work okay on my end after running and comparing the output with that of the master branch using:

  `python toml_test.py < examples/example-v0.4.0.toml`

It could be the case that the PR completely crashes and burns when running the tests properly. Let me konw if that's the case as I might need some guidance running them correctly.

Cheers, Jack

### DIFF
--- a/toml.py
+++ b/toml.py
@@ -781,15 +781,20 @@ def _dump_inline_table(section):
         return str(_dump_value(section))
 
 def _dump_value(v):
-    dump_value = {
-        str: _dump_str(v),
-        unicode: _dump_str(v),
-        list: _dump_list(v),
+    dump_funcs = {
+        str: lambda: _dump_str(v),
+        unicode: lambda: _dump_str(v),
+        list: lambda: _dump_list(v),
         bool: lambda: str(v).lower(),
         float: lambda: str(v),
         datetime.datetime: lambda: v.isoformat()[:19]+'Z',
     }
-    return dump_value.get(type(v), v)
+    # Lookup function corresponding to v's type
+    dump_fn = dump_funcs.get(type(v))
+    # Evaluate function (if it exists) else return v
+    return dump_fn() if dump_fn is not None else v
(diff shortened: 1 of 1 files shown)


toml.py
```diff
@@ -696,38 +696,45 @@ def _dump_sections(o, sup):
...
-        v = v.replace("\\x", "\\u00")
-        return str('"'+v+'"')
-    if isinstance(v, bool):
-        return str(v).lower()
-    if isinstance(v, datetime.datetime):
-        return v.isoformat()[:19]+'Z'
-    if isinstance(v, float):
-        return str(v)
-    return v
+    dump_value = {
```
uiri: The way that this is constructed, it calls every single possibility on `v` when it should only evaluate the matching possibility. I think that `dump_value` should map types to functions and the `return` statement below should call that function on `v` (with the default being an identity lambda, something like `lambda x: x`).

uiri:
Hi Jack,

Thank you for the pull request.

Testing for the repository is something which I am working on improving. The `toml_test.py` script is actually a harness for a language agnostic TOML testing library (which is, unfortunately, not as well maintained as I would hope). Automatic Travis CI has been added since you submitted your pull request. I am actively looking at ways to improve test coverage. I hope that the situation will be better in a month or two when I have a little bit more time to devote to improving the library.

While performing some manual testing of your patch, I ran into stack overflows. I believe that they are due to the issue I've pointed out via the code comment above.

-Will

uiri:
Thanks again for the PR! I've rebased and merged it.

PR #97, opened 2017-05-06
Enable support for toml.__version__

mottosso:
Sometimes it can be useful to reference what version of `toml` one is using, especially as the spec continues to change.

```python
>>> import toml
>>> print(toml.__version__)
'0.9.2'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,10 @@
+import toml
 from distutils.core import setup
 
 with open("README.rst") as readmefile:
     readme = readmefile.read()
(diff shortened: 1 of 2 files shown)


uiri:
Hi Marcus,

Thanks for the PR.

Do you think making it easy to reference the relevant version of the spec as `toml.__spec__` would also be useful? Currently, the information is in the README.

mottosso:
The thought did occur to me, and I would agree. Adding it now.

Issue #9, opened 2013-05-13
Option to load many config files


matrixik:
I like from Python ConfigParser option to load many config files:

``` python
from ConfigParser import SafeConfigParser

parser = SafeConfigParser()

parser.read(['does_not_exist.ini', 'config.ini', 'config_dev.ini', ])

print 'Log level:', parser.get('logging', 'log_level')
print 'Prefix:', parser.get('logging', 'prefix')
```

if `'config.ini`:

``` ini
[logging]
log_level = info
prefix = 'my_prefix'
```

and `config_dev.ini`:

``` ini
[logging]
log_level = debug
```

Output would be:

```
Log level: debug
Prefix: "my_prefix"
```

If I remove `config_dev.ini`, output will be:

```
Log level: info
Prefix: "my_prefix"
```

> ConfigParser overwrites the keys with each successive file, the order in which the files are read is determined by the order of the file names in the list passed to ConfigParser.read

Docs: http://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.read

If files not exists instead of throwing error:

> If none of the named files exist, the ConfigParser instance will contain an empty dataset.

From http://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.read

Would be nice if you could add something similar.

Best regards,
Dobrosław Żybort

P.S. Now, to merge multi-level dicts, I need to use http://stackoverflow.com/a/10704003/1722542

machielg:
Could this be a regression? I don't see any reference to the fix for this issue in the current code. It looks like the [current version](https://github.com/uiri/toml/blob/230f0c9cffd230354759949dd1dd89acd3dddb80/toml/decoder.py#L149) uses the python update which breaks the initially requested behaviour. At least in my own code I see a simple key overwriting happening in v0.10.2

PMeira:
@machielg I was curious about this but couldn't find any explanation why it was removed besides the commit message from ca3d32e574f85e364d104117af0a6c7fa667910d, from 2016.

Unless I missed a note somewhere, maybe a note about this incomplete merge behavior could be added, especially since the module did use a different merge strategy in the past.

machielg:
Can we reopen this issue? Or a new one?

PR #99, opened 2017-05-08
Feature: Let load() handle non-existing files in list.

x10an14:
As long as there's at least _one_ existing file in the list given.
To permit things like the examples listed here: https://github.com/uiri/toml/issues/9#issuecomment-17939193

Hopefully the result becomes either what's in `config.ini` or `config_dev.ini` if either's opposite is non-existent on the filesystem.

### DIFF
--- a/toml.py
+++ b/toml.py
@@ -64,9 +64,10 @@ def load(f, _dict=dict):
         Parsed toml file represented as a dictionary
 
     Raises:
-        FileNotFoundError -- When array with no existing file path(s) is passed
         TypeError -- When f is invalid type
         TomlDecodeError: Error while decoding toml
+        IOError -- When array with zero valid (existing) file paths is passed (Python 2)
+        FileNotFoundError -- When array with zero valid (existing) file paths is passed (Python 3)
     """
 
     if isinstance(f, basestring):
@@ -79,7 +80,11 @@ def load(f, _dict=dict):
             error_msg = "Load expects a list to contain filenames only."
             error_msg += linesep
             error_msg += "The list needs to contain the path of at least one existing file."
-            raise FileNotFoundError(error_msg)
+            try:
+                raise FileNotFoundError(error_msg)
+            except NameError:
+                # Program executed with Python 2, not Python 3:
+                raise IOError(error_msg)
         d = _dict()
         for l in f:
             d.update(load(l))
--- a/toml.py
+++ b/toml.py
@@ -68,9 +68,17 @@ def load(f, _dict=dict):
         with io.open(f, encoding='utf-8') as ffile:
             return loads(ffile.read(), _dict)
     elif isinstance(f, list):
-        for l in f:
-            if not isinstance(l, basestring):
-                raise TypeError("Load expects a list to contain filenames only")
+        from os import path as op
+        # Should we remove erroneous entries?
+        #   Say, permit non-existing files?
+        # Then comment in below line:
+        # f = [path for path in f if op.exist(path)]
+        if not any(op.exist(l) for l in f):
+            from os import linesep
+            error_msg = "Load expects a list to contain filenames only."
+            error_msg += linesep
+            error_msg += "The list needs to contain the path of at least one existing file."
+            raise TypeError(error_msg)
         d = _dict()
         for l in f:
             d.update(load(l))
--- a/toml.py
+++ b/toml.py
@@ -1,8 +1,10 @@
 # This software is released under the MIT license
 
 import re
-import datetime
 import io
+import datetime
+from os import linesep
+
 
 __version__ = "0.9.2"
 __spec__ = "0.4.0"
@@ -62,7 +64,7 @@ def load(f, _dict=dict):
         Parsed toml file represented as a dictionary
 
     Raises:
-        TypeError -- When array of non-strings is passed
+        FileNotFoundError -- When array with no existing file path(s) is passed
         TypeError -- When f is invalid type
         TomlDecodeError: Error while decoding toml
     """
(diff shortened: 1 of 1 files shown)


uiri:
Hi x10an14,

Thank you for your pull request! I am sorry it took me a little while to take a proper look at it.

It looks like you made a typo and wrote `exist` instead of `exists`. I would prefer `path.exists` to `op.exists` but it is a matter of personal preference.

In the case of a list with some existent and some nonexistent filenames, the `load` still raises an error on the nonexistent files. I think it should just ignore the nonexistent filenames and error out if there are no existent filenames in the list. You already wrote the code, it is just commented out :)

I don't think a lack of valid filenames should be a TypeError since all the items in the list are the right type. I am leaning towards a ValueError or an IOError instead.

I think linesep might be useful elsewhere in the code. In which case, importing it in a larger scope would make sense.

Thanks again!

x10an14:
Hi @uiri! =)

Thanks for the feedback! And don't worry about the promptness =).
Let me respond to your feedback point-by-point:

- > It looks like you made a typo and wrote `exist` instead of `exists`.

Sorry! My bad! I'll check why my linter didn't catch that. In any case, I probably ought to re-run the tests, which I thought I had done...

- > I would prefer `path.exists` to `op.exists` but it is a matter of personal preference.

Yeah, like you say, personal preference =). Feel free to revert/change that if you so choose!
My reasoning for using `op` instead of `path`:

1. Any `from x import y` is (for me) a shorthand to save on horizontal space, so you don't need to write `x.y` everywhere (which can become long in our case: `os.path.exists`). This I find especially important when trying to respect a maximum horizontal line length of 79/80/119/120.
Imagine the case where you want to get the real and absolute filepath of something:
`os.path.abspath(os.path.realpath(os.path.join(some_path, some_file))))` vs `op.abspath(op.realpath(op.join(some_path, some_file)))`
2. I 'chose' the `os` shorthand in that regard, with the little logic that it's from the `os` module, and the `path` submodule. But again, I won't feel slighted at all if you prefer to change it for your own readability!

- > In the case of a list with some existent and some nonexistent filenames, the `load` still raises an error on the nonexistent files.

'Twas intentional in the hope to raise some discussion about the behaviour you'd prefer! =D
  * >  I think it should just ignore the nonexistent filenames and error out if there are no existent filenames in the list. You already wrote the code, it is just commented out :)

Done! =)

- > I don't think a lack of valid filenames should be a TypeError since all the items in the list are the right type. I am leaning towards a ValueError or an IOError instead.

I was wondering the same, but I wanted your input (again, trying to provoke discussion by leaving it be^^).
Personally I find it more "Pythonic" with `ValueError`, but more 'correctly descriptive' with `IOError`.
How'd you feel about `FileNotFoundError`?

- > I think linesep might be useful elsewhere in the code. In which case, importing it in a larger scope would make sense.

I'll fix that in a separate commit! I'll see what I can find =) Maybe just aswell a different pull-request?

x10an14:
I found out why I didn't run your tests. I didn't figure out how to!

I found your file [TESTING](https://github.com/uiri/toml/blob/3869c1dd250dfcce98fa8acdaac06c25a5b3e159/TESTING), and I installed Go, managed to execute its 'default tests', but that's as far as I got =/.

See [my gist](https://gist.github.com/x10an14/5fc87eb45e1ec6564ecb3a083bf095e0) for my attempt!

x10an14:
@uiri I also see now in your [.travis.yml](https://github.com/uiri/toml/blob/dab00e9880506bc48ec3575f66f5b6ea6f82e998/.travis.yml) that you actually don't use these `toml_test.py`/`toml_test3.py`. 

Would you mind if I make another PR for updating your testing instructions? =)

uiri:
I am sorry that the Testing documentation is lagging behind the actual code and operations. I have more improvements I want to make to the testing. The most significant one is adding tests for all the resolved issues to catch regressions.

> - > It looks like you made a typo and wrote `exist` instead of `exists`.
> 
> Sorry! My bad! I'll check why my linter didn't catch that. In any case, I probably ought to re-run the tests, which I thought I had done...

The tests don't execute that code path. They test language agnostic encoding and decoding thoroughly. The coverage for the actual python API is scarce.
 
> - > In the case of a list with some existent and some nonexistent filenames, the `load` still raises an error on the nonexistent files.
> 
> 'Twas intentional in the hope to raise some discussion about the behaviour you'd prefer! =D

I think perhaps there should still be a warning that some of the files were not found. I can look into that if you are unfamiliar with Python's warning facilities.

> I was wondering the same, but I wanted your input (again, trying to provoke discussion by leaving it be^^).
> Personally I find it more "Pythonic" with `ValueError`, but more 'correctly descriptive' with `IOError`.
> How'd you feel about `FileNotFoundError`?

FileNotFoundError does not exist in Python 2. I think IOError for Python 2 and FileNotFoundError for Python 3 makes the most sense. FileNotFoundError only exists in version Python 3.3 or later but that is fine since we are not targeting earlier Python 3 versions.

Thanks again for your PR and muddling through the poorly documented testing procedures.

x10an14:
> I am sorry that the Testing documentation is lagging behind the actual code and operations. I have more improvements I want to make to the testing. 

No worries =).
Didn't mean to call you out on anything lacking, would be happy to prepare my PRs better if you don't mind me communicating with you over email.

> The tests don't execute that code path. They test language agnostic encoding and decoding thoroughly. The coverage for the actual python API is scarce.

Ahh okay. 
As you might've gathered, I'm not that good/into Go. 
My interest lies primarily in Python as opposed to Go. 
I don't mind leaving the actual TOML decoding tests to your Go suite or something else, but I could perhaps be interested in improving the Python coverage (at least for Python3).

> I think perhaps there should still be a warning that some of the files were not found. I can look into that if you are unfamiliar with Python's warning facilities.

I must admit I had never heard of the `import warning` module before you mentioned it here. 
I'm not quite sure how you'd like it's implementation, since I'm getting the feeling that our programming styles diverge some.
So if you wouldn't mind giving some examples/commit(s) to this branch, then I could take it from there if you'd like!
(I must admit I was wondering how to send execution updates to the users through stderr, and this seems like the 'correct' way to do it, although I must admit I like [this S/O answer](https://stackoverflow.com/a/14981125)).

> FileNotFoundError does not exist in Python 2. I think IOError for Python 2 and FileNotFoundError for Python 3 makes the most sense. FileNotFoundError only exists in version Python 3.3 or later but that is fine since we are not targeting earlier Python 3 versions.

Alright!
Commit incoming!

> Thanks again for your PR and muddling through the poorly documented testing procedures.

No worries & my pleasure =). 
Please don't hesitate to call me out on anything if there's something you dislike/would like to discuss about my PRs.

uiri:
This looks good to me. Merge'd.

> [...]  would be happy to prepare my PRs better if you don't mind me communicating with you over email.

Feel free. My email is in the commit logs for a reason.

I opened up #104 to track the warning issue. I haven't used `import warning` before either so I'm afraid I don't have any ready examples to point to.

PR #101, opened 2017-05-17
Pep8ify the source code

x10an14:
Hi! 

In PR #99, there was some discussing regarding the use of `os.linesep` instead of manually looking for/replacing `\r\n` or `\n`.

As such, I'd like to PEP8ify the code base in [toml.py](https://github.com/uiri/toml/blob/master/toml.py) before I come with such a PR.
(It would make it easier to conform to a standard of code quality, plus help the refactoring quite a bit in my experience).

This PR thus contains:

- A [.flake8](https://github.com/x10an14/toml/blob/cc6a2dc8d8558c7d2c75c00b7b1eb894b9d1efae/.flake8) config file specifying the PEP8 rules I'd suggest this project use (which is absolutely open for debate!).
- A `toml.py` file edited to live up to the rules specified in `.flake8`, with one exception, as stated in this [commit](https://github.com/x10an14/toml/commit/cc6a2dc8d8558c7d2c75c00b7b1eb894b9d1efae#diff-540ed609f086a01ee6f84e1a7bba5f6e).

This PR is also based off on PR #99, and I'll rebase this on top of that one as required if there's anything more to be done in #99. =)

PR #102, opened 2017-05-17
Feature: Updated TESTING instructions.

x10an14:
Future TODO: Make .py test-files executable without polluting machine environment.
(AKA: Through a virtualenv permit `import toml` without raising `ImportError`).

(This PR is independent of both PRs #99 and #101).


uiri:
Thank you for re-writing this!

The test suite that should be used is my fork - https://github.com/uiri/toml-test
The BurntSushi test suite is missing approximately 30 tests.

The harness script has been relocated to `tests/decoding_test.py`.

There used to be a separate harness for python3, however that has been relocated to `tests/example_test.py`. I believe this script can be safely removed.

This toml library is intended to be compatible with both python2 and python3. The test suite should therefore be run against both versions. Whichever python2 and python3 one has installed locally ought to be fine - continuous integration takes care of a more thorough round of testing against various releases of each version. The `decoding_test.sh` is intended for Travis while `decoding_test2.sh` and `decoding_test3.sh` is intended for local developer use.

I'll also note that `print()` is valid in both Python 2 and Python 3. `print x` is valid in Python 2 but invalid in Python 3.

How would you feel about renaming this file to CONTRIBUTING? Is there any other information which you think might be helpful to new contributors? I have been working on this code base on and off for so long that it is a little bit difficult for me to see the forest for the trees without a bit of prodding.

I would prefer the file not have the `md` extension. I think the filename looks cleaner without an extension. The license file does not have an extension either and the README only has an extension so for pretty printing on PyPI.

PR #111, opened 2017-06-28
Improve project documentation and convert to Setuptools

tmose1106:
# Commit Mesage

I original created this commit because I received a warning from pip
about the project using the deprecated distutils module. So I converted
the setup.py file to Setuptools, but since it is simple enough, I was
able to retain Distutils compatability using a try statement.

While I was at it, I decided to overhaul the CONTIRUBUTING and README
files by improving their usage of reStructuredText. They now appear less
cluttered and more organized. Similarly, I added extensions to the
CONTRIBUTING and LICENSE files, since all files should really have some
kind of extension.

# Message to Maintainer

I have had success with you project thus far, and I have decided to add the
changes discussed above. They only deal with the documentation of the
project, and nothing really code related, so it should all be good to go. If
you have any further requests prior to accepting this pull request, I am open
to any suggestions. Thanks for your hard work.


uiri:
Hi Ted,

Apologies for the late response.

Thank you for your documentation and distribution efforts.

I think that LICENSE should stay without a file extension. That seems to be the norm in open source projects (e.g. CPython itself at github.com/python/cpython). File extensions are an MS-DOS/Windows convention so my UNIX bias is definitely showing through here.

I am a little concerned that GitHub won't recognize CONTRIBUTING.rst while it would recognize CONTRIBUTING or CONTRIBUTING.md. I do recognize that mixing both reStructuredText and markdown in the repository is a little awkward.

Everything else looks good; just those two nits.

tmose1106:
I too am an avid GNU/Linux user, but I personally think that putting extensions on all files (besides executables) is especially important for code repositories since many tools (on all OSes) use a file's extension to gather information about it. My personal bias comes about due to the fact that I have a project called [Metrics](https://gitlab.com/tmose1106/metrics) which calculates the SLOC, number of comments, etc within a file or project, and it determines the type of a file based upon its file extension. Nevertheless, this is your project, and I respect your opinion on the subject.

As for the contributing guide, I just sent an email to GitHub Support on the subject, so hopefully that issue will be rectified within the near future.

Finally, would you like me to handle your two requests in a new commit?

uiri:
>  I too am an avid GNU/Linux user, but I personally think that putting extensions on all files (besides executables) is especially important for code repositories since many tools (on all OSes) use a file's extension to gather information about it. My personal bias comes about due to the fact that I have a project called [Metrics](https://gitlab.com/tmose1106/metrics) which calculates the SLOC, number of comments, etc within a file or project, and it determines the type of a file based upon its file extension. Nevertheless, this is your project, and I respect your opinion on the subject.

You may be interested in looking at how the `file` command does it. http://www.darwinsys.com/file/
 
> As for the contributing guide, I just sent an email to GitHub Support on the subject, so hopefully that issue will be rectified within the near future.

Thanks!

> Finally, would you like me to handle your two requests in a new commit?

Yes, a new commit would be the best way to update the pull request.

PR #128, opened 2017-10-02
Fix #100: track position while parsing for better errors

### LINKED ISSUE #100: TomlDecodeError exception should include file position
Things like "Duplicate keys" would be easier for end user to track down if the message included file char number, or line/column where detected.  This could be part of the exception string or a member on the exception class.

pdedmon:
I made the `TomlDecodeError` more like the `JSONDecodeError` from the stdlib json module, allowing position information to be surfaced in errors.

This partially addresses https://github.com/uiri/toml/issues/100, as line numbers are reported accurately in errors. However, column numbers are harder to track in the current code; as a result, errors will often say "column 1", when the issue occurred elsewhere in the line. Also, errors in multiline arrays report the opening line, because they are currently in-lined before parsing.

### DIFF
--- a/toml/decoder.py
+++ b/toml/decoder.py
@@ -445,8 +445,8 @@ def _load_unicode_escapes(v, hexbytes, prefix):
         while i < hxblen:
             try:
                 if not hx[i].lower() in hexchars:
-                    raise ValueError()
-            except (IndexError, ValueError):
+                    raise ValueError("Invalid hex character")
+            except IndexError:
                 raise ValueError("Invalid escape sequence")
             hxb += hx[i].lower()
             i += 1
--- a/toml/decoder.py
+++ b/toml/decoder.py
@@ -20,9 +20,19 @@ except NameError:
     FNFError = IOError
 
 
-class TomlDecodeError(Exception):
+class TomlDecodeError(ValueError):
     """Base toml Exception / Error."""
-    pass
+
+    def __init__(self, msg, doc, pos):
+        lineno = doc.count('\n', 0, pos) + 1
+        colno = pos - doc.rfind('\n', 0, pos)
+        emsg = '{} (line {} column {} char {})'.format(msg, lineno, colno, pos)
+        ValueError.__init__(self, emsg)
+        self.msg = msg
+        self.doc = doc
+        self.pos = pos
+        self.lineno = lineno
+        self.colno = colno
 
 
 # Matches a TOML number, which allows underscores for readability
@@ -129,6 +139,7 @@ def loads(s, decoder=None):
     if not isinstance(s, unicode):
         s = s.decode('utf8')
 
+    original = s
     sl = list(s)
     openarr = 0
     openstring = False
@@ -145,7 +156,7 @@ def loads(s, decoder=None):
         if keyname:
             if item == '\n':
                 raise TomlDecodeError("Key name found without value."
-                                      " Reached end of line.")
+                                      " Reached end of line.", original, i)
             if openstring:
                 if item == openstrchar:
                     keyname = 2
@@ -164,7 +175,8 @@ def loads(s, decoder=None):
                 keyname = 0
             else:
                 raise TomlDecodeError("Found invalid character in key name: '" +
-                                      item + "'. Try quoting the key name.")
+                                      item + "'. Try quoting the key name.",
+                                      original, i)
         if item == "'" and openstrchar != '"':
             k = 1
             try:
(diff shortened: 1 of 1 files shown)


uiri:
Hi pdedmon,

The Travis failure is due to some lines exceeding 80 characters. Not a huge deal. I have a pre-commit hook I use to warn me about these things but I am not sure what is the best way to share it in the repository.

My initial thoughts with regards to the issues tagged `user customization` were to allow them to be fixed by refactoring the code into a `Decoder` and `Encoder` model similar to that used by the python standard library for JSON.

I will take a closer look at this PR this week to see how I feel about your proposed solution. It may be easier to fix #100 and #77 together since they both will rely on the unmangled `toml` source.

pdedmon:
No worries! I just noticed the `.flake8` file in the repo - I'll make sure to run `flake8` before committing next time.

+1 for the `Encoder/Decoder` model!

gaborbernat:
@pdedmon any updates on this?

pdedmon:
@gaborbernat I'd forgotten about this actually, but I might look into the conflicts and build failures later.

pdedmon:
I rebased and did a quick manual test (on top of the toml-test and tox tests), and it seems to be working as before.

gaborbernat:
@uiri does this look good?

toml/decoder.py
```diff
@@ -417,9 +445,9 @@ def _load_unicode_escapes(v, hexbytes, prefix):
         while i < hxblen:
             try:
                 if not hx[i].lower() in hexchars:
-                    raise IndexError("This is a hack")
-            except IndexError:
-                raise TomlDecodeError("Invalid escape sequence")
+                    raise ValueError()
+            except (IndexError, ValueError):
```
uiri: I think this should still be an `IndexError`? Why the change to `ValueError`?

uiri:
Hi,

I think that the `TomlDecodeError`s should _not_ change to `ValueError`s. I think it would be better to have an alternative `__init__` for the `TomlDecodeError` for the case where there is no file position available.

What is the reasoning behind changing to `ValueError` ?

toml/decoder.py
pdedmon: I thought the index error was a hack to fall into this `except`, and that a `ValueError` was more appropriate for an invalid hex character.

pdedmon:
I inherited from `ValueError` because that's what the python stdlib does with JSONDecodeError: https://docs.python.org/3/library/json.html#json.JSONDecodeError

uiri:
Inheriting from `ValueError` isn't the problem. I think that that makes a lot of sense, actually!

I see that there are a lot of instances of `TomlDecodeError` which this patch replaces with `ValueError`. I think it would be better to clearly indicate that an error from the library during decoding came from the library and not some other part of a program which happens to consume the library.

toml/decoder.py
uiri: I suppose you are right. It doesn't really matter (to me) _which_ error type is used but I think the `except` should only accept the intentional error which gets raised for an invalid hex character.

gaborbernat:
@uiri so what changes do you request/propose?

pdedmon:
I replaced the uses of TomlDecodeError with ValueError in some of the functions, because those functions didn't have access to the decoder position information. I suppose you could continue to raise a TomlDecodeError without any information, and then augment it before re-raising, but using a ValueError made more sense to me. All of the ValueErrors that have replaced TomlDecodeErrors are caught and result in TomlDecodeErrors, so we are not letting the ValueErrors escape from the lib.

uiri:
Oh! I see, now, thanks for pointing out the `except ValueError` clauses which you added. Somehow that didn't click for me before.

So then I think it is just the nit around the (slightly, overly) broad `except (IndexError, ValueError):` line which I would prefer narrowed.

Thank you again for your PR and for doing the rebase after the recent significant changes to master.

pdedmon:
No worries! I'll update that `except` line now.

toml/decoder.py
```diff
@@ -445,8 +445,8 @@ def _load_unicode_escapes(v, hexbytes, prefix):
         while i < hxblen:
             try:
                 if not hx[i].lower() in hexchars:
-                    raise ValueError()
-            except (IndexError, ValueError):
+                    raise ValueError("Invalid hex character")
+            except IndexError:
```
uiri: Uhhh... shouldn't the raised error match the excepted error?

pdedmon:
Not necessarily, why?

Issue #130, opened 2017-10-05
dumps(preserve=True) does not enquote keys with whitespace


NiklasRosenstein:
```python
>>> import toml
>>> d = toml.loads('[authors]\n"Niklas Rosenstein" = { email = "rosensteinniklas@gmail.com" }')
>>> print(d)
{'authors': {'Niklas Rosenstein': {'email': 'rosensteinniklas@gmail.com'}}}
>>> print(toml.dumps(d, preserve=True))
[authors]
Niklas Rosenstein = { email = "rosensteinniklas@gmail.com" }
```

The resulting dump is invalid TOML. v0.9.3

NiklasRosenstein:
See PR #131

PR #131, opened 2017-10-05
fix inline table formatting (close uiri/toml#130)

NiklasRosenstein:


uiri:
Thank you Niklas for finding and fixing this bug!

PR #132, opened 2017-10-05
add *args, **kwargs to dump() and forward to dumps()

NiklasRosenstein:


uiri:
Hi Niklas,

Thank  you for your pull request! What exactly does this achieve? While I am vaguely familiar with `args` and `kwargs` and their usage, I am not sure what the point of adding them to `dump` and `dumps` is.

NiklasRosenstein:
It allows you to pass pass the `preserve` argument to `toml.dump()`. While this could also be achieved without `*args, **kwargs` by simply adding `preserve=False` as an argument to `dump()`, this way will ensure that any arguments added to `dumps()` (note the S) in the future will automatically be available to `dump()` as well.

uiri:
I have no immediate plans to add any more arguments to `dumps` nor to `loads`. There are a number of issues currently open which I want to address in other ways since the list of arguments is starting to get a bit unwieldy and some of the requested modifications are going to be overkill for the vast majority of users. I think this might be overkill.

That said, if this is the right move:

1. Should `*args, **kwargs` also encompass the `_dict` argument?
2. Does this make sense for `load` / `loads` too?

NiklasRosenstein:
It would certainly make sense, although I would say that it's a matter of taste. Using `*args, **kwargs` is convenient, but less verbose and less self-documenting than repeating the argument list.

uiri:
Hi Niklas,

I am closing this PR since the switch to an Encoder/Decoder model has supplanted the need for it.

Please feel free to reopen it if you still believe it is necessary.

Thank you again for the PR!

PR #136, opened 2017-10-22
New API: Encoder / Decoder model

uiri:
Firstly, I am sorry that git's diff algorithm can't understand the changes in this PR for the life of it. Even in split mode, it looks far messier than it actually is. See #135 for context around splitting up the module into a package of modules (i.e. what the hell happened in the base that this PR is `diff`'d against).

So the 0.9.X Toml API is something like:

`dumps(o, preserve=False)`
`loads(s, _dict=dict)`

The new Toml API I'm proposing is something like:

`dumps(o, decoder=None)`
`loads(s, encoder=None)`

Given a `None` encoder or decoder, the method creates one from the default `TomlEncoder` or `TomlDecoder` class. Examples of custom Encoders are: `TomlArraySeparatorEncoder`, `TomlPreserveInlineDictEncoder`, and the `TomlOrderedEncoder`. An example of a custom Decoder is `TomlOrderedDecoder`.

The `TomlArraySeparatorEncoder` seems kind of clumsy but I am not sure which facility is general purpose enough to expose while also accomplishing the desired goal of custom newlines/spacing around encoded array elements.

So, I want to minimize breaking downstream users of the library. This model has an inherent flexibility which avoids future breakage by containing it within the Decoder/Encoder class. Of course, the APIs presented by those classes could change. One strategy to avoid future breakages caused by *those* changes is to include a set of custom Encoders/Decoders with the `toml` library.

To that end, I am soliciting feedback on these changes and the overall new API design.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -26,6 +26,14 @@ def test_valid_tests():
         toml.dumps(toml.load(open(os.path.join(valid_dir, f))))
 
 
+def test__dict():
+    class TestDict(dict):
+        pass
+
+    assert isinstance(toml.loads(
+        TEST_STR, _dict=TestDict), TestDict)
+
+
 def test_dict_decoder():
     class TestDict(dict):
         pass
--- a/toml/__init__.py
+++ b/toml/__init__.py
@@ -12,8 +12,9 @@ __spec__ = "0.4.0"
 load = decoder.load
 loads = decoder.loads
 TomlDecoder = decoder.TomlDecoder
+TomlDecodeError = decoder.TomlDecodeError
 
 dump = encoder.dump
 dumps = encoder.dumps
-
-TomlDecodeError = decoder.TomlDecodeError
+TomlEncoder = encoder.TomlEncoder
+TomlArraySeparatorEncoder = encoder.TomlArraySeparatorEncoder
--- a/toml/__init__.py
+++ b/toml/__init__.py
@@ -6,7 +6,7 @@ Released under the MIT license.
 from toml import encoder
 from toml import decoder
 
-__version__ = "0.9.3"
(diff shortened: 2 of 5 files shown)


samvasko:
Could we keep the current API and add the `decoder=` or `encoder=`.  In the background the decoder and encoder would be used for all the options.

uiri:
So one approach is to generate a `warning` on the old syntax for now and then fully move to dropping those parameters in a future release (most likely the one after given the current length of time between releases).

Another approach - which I am leaning more towards - is to have a `legacy` (or whatever... name is a bit arbitrary) module which maintains the old API for at least a little while. So downstream users could change:

`import toml`

to

`from toml import legacy as toml`

And not touch any of the rest of their code, if they so choose.

samvasko:
Could you please explain the reasoning behind removing the old syntax? It is much easier to type and understand a `preserve=True` than `encoder=TomlPreserveInlineDictEncoder` or similar. I also this the background change warrant a breaking change in the API

uiri:
`preserve` becomes ambiguous when that could mean preserve inline dictionaries or preserve comments. (There could potentially also be "preserve whitespace" but that seems a tad overboard to me.)

If I was designing the API today from first principles without any regard to backwards compatibility, I would definitely remove the `preserve` flag. After thinking about it some more, I could be convinced to keep the `_dict` flag, but it depends on how difficult keeping it would make integration with the `encoder`/`decoder` API. Encapsulating everything in the `encoder`/`decoder` simplifies the handling inside the external methods themselves even if it ultimately just pushes the complexity elsewhere rather than eliminate it.

A more extensive refactor would more easily justify breaking backwards compatibility. I will play around with it and see whether or not I like the results.

PR #141, opened 2017-11-07
Refactor testing

samvasko:
Some tooling improvements! 

- Converted tests to pytest
- Added tox as a easy way to do multi python testing locally
- Expanded gitignore
- Added coverge using codecov @uiri you might need to set it up for this repo

🐎  Ready for review 🏇


.gitignore
```diff
@@ -1,12 +1,104 @@
-# Filetypes
-*~
```
uiri: Can you re-add this ignore for the "backup" files that Emacs (my editor of choice) creates?

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: python
 python:
- - "2.6"
```
uiri: There is (at least) one downstream project which maintains 2.6 compatibility which I'd rather not break.

tests/test_api.py
```diff
@@ -0,0 +1,60 @@
+import toml
```
uiri: > Rename api_test.py to test_api.py

😂 👍

CONTRIBUTING
```diff
@@ -15,6 +15,25 @@ latest release.
 Testing
 =======
 
+### Unit tests
+Unit tests can be run using [tox](https://tox.readthedocs.io/en/latest/).
+Simply `pip install tox` and you are ready to go. Tox creates requried
```
uiri: Typo: requried → required

.gitignore
samvasko: I can but then should I add vim and Sublime Text ignore files? I recommend adding the emacs backup files into your [global .gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)

.travis.yml
samvasko: Python 2.6 is officially not supported. Are you sure this is a good thing to spend time on?

.gitignore
uiri: I'm OK with having multiple editors' worth of crud in the `.gitignore`.

uiri:
LGTM :+1:

PR #157, opened 2018-04-14
Include type hints so they're shipped in the wheel

jrabbit:
looks ready to go otherwise


uiri:
No objections from me. Merge'd.

PR #162, opened 2018-05-08
Add sample output

ocsw:
It was super hard to find what toml output looked like, so I ended up running .load() on the example file.


uiri:
Thanks! LGTM

PR #175, opened 2018-07-14
Add support for hexadecimal, octal and binary numbers

Fieoner:
Now this works
```
> import toml
> a="var = 0x100"
> toml.loads(a)
{u'var': 256}
```
As well as this:
```
> import toml
> a = """
... hex = 0xAB03
... binary = 0b100101
... octal = 0o630
... """
> toml.loads(a)
{u'octal': 408, u'binary': 37, u'hex': 43779}
```
acording to [0.5.0 spec](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.5.0.md#user-content-integer)

> Non-negative integer values may also be expressed in hexadecimal, octal, or binary.

### DIFF
--- a/toml/decoder.py
+++ b/toml/decoder.py
@@ -703,7 +703,7 @@ class TomlDecoder(object):
             elif v[0] == '+':
                 v = v[1:]
             v = v.replace('_', '')
-            if v[1] not in ['x', 'o', 'b']:
+            if 'x' not in v:
                 if '.' in v or 'e' in v or 'E' in v:
                     if '.' in v and v.split('.', 1)[1] == '':
                         raise ValueError("This float is missing digits after "
@@ -713,7 +713,7 @@ class TomlDecoder(object):
                                          "digit")
                     v = float(v)
                     itype = "float"
-            else:
+            if itype == "int":
                 v = int(v, 0)
             if neg:
                 return (0 - v, itype)
--- a/toml/decoder.py
+++ b/toml/decoder.py
@@ -75,8 +75,12 @@ def _strictly_valid_num(n):
         return False
     if len(n) == 1:
         return True
-    if n[0] == '0' and n[1] != '.':
-        return False
+    if n[0] == '0':
+        if n[1] in ['o','b','x']:
+            n = int(n, 0)
+            return True
+        if n[1] != '.':
+            return False
     if n[0] == '+' or n[0] == '-':
         n = n[1:]
         if n[0] == '0' and n[1] != '.':
@@ -712,7 +716,7 @@ class TomlDecoder(object):
                 v = float(v)
                 itype = "float"
             else:
-                v = int(v)
+                v = int(v, 0)
             if neg:
                 return (0 - v, itype)
             return (v, itype)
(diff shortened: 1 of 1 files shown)


toml/decoder.py
```diff
@@ -75,8 +75,12 @@ def _strictly_valid_num(n):
...
     if len(n) == 1:
         return True
-    if n[0] == '0' and n[1] != '.':
-        return False
+    if n[0] == '0':
+        if n[1] in ['o','b','x']:
+            n = int(n, 0)
+            return True
+        if n[1] != '.':
+            return False
```
uiri: I think this should just be `if n[0] == '0' and n[1] not in ['.','o','b','x']` then `return False`
I am concerned that `int(n, 0)` might cause a ValueError to bubble up where it shouldn't.

uiri:
Hi César,

Thanks for the pull request! Can you fix the style (spacing) issues which are making the tests fail? I know it is a bit of a pain haha.

toml/decoder.py
Fieoner: I also agree that my approach looks a bit ugly but I wanted to avoid passing floats to int(n, 0).

Fieoner: Forget about this ^ I just realised how stupid my comment is

toml/decoder.py
```diff
@@ -702,17 +703,18 @@ def load_value(self, v, strictly_valid=True):
...
-                    raise ValueError("This float is missing digits after "
-                                     "the point")
-                if v[0] not in '0123456789':
-                    raise ValueError("This float doesn't have a leading "
-                                     "digit")
-                v = float(v)
-                itype = "float"
-            else:
-                v = int(v)
+            if 'x' not in v:
```
uiri: Why 'x' and not 'o' or 'b' ?

toml/decoder.py
```diff
@@ -702,17 +703,18 @@ def load_value(self, v, strictly_valid=True):
...
+                if '.' in v or 'e' in v or 'E' in v:
+                    if '.' in v and v.split('.', 1)[1] == '':
+                        raise ValueError("This float is missing digits after "
+                                         "the point")
+                    if v[0] not in '0123456789':
+                        raise ValueError("This float doesn't have a leading "
+                                         "digit")
+                    v = float(v)
+                    itype = "float"
+            if itype == "int":
```
uiri: Where does itype get set to `"int"` ?

toml/decoder.py
Fieoner: Next line is 
```if '.' in v or 'e' in v or 'E' in v:```
so hex numbers containing 'e' will be mistaken as a float. That is not a problem with binary and octal numbers so only 'x' is needed there

toml/decoder.py
Fieoner: Line 697
But it will be changed to "float" if the number is, well, a float.
Base 10 numbers are fine without this but hex, binary and octal need to be converted to base 10 with the ```v = int(v, 0)``` thing in the next line.

uiri:
Thanks again for the PR! Hopefully this can be the _second_ library to support v0.5.0 :)

PR #179, opened 2018-08-07
Added support for encoding of tuples.

jkarns275:
Right now tuples will are encoded using a to string call (I think - that is the behavior I observed), this just makes tuples get encoded as lists, so they can be decoded.

### DIFF
--- a/toml/encoder.py
+++ b/toml/encoder.py
@@ -116,6 +116,7 @@ class TomlEncoder(object):
             str: _dump_str,
             unicode: _dump_str,
             list: self.dump_list,
+            tuple: self.dump_list,
             bool: lambda v: unicode(v).lower(),
             int: lambda v: v,
             float: _dump_float,


uiri:
Hi Josh,

Thank you for your pull request! I believe that [the check on line 158](https://github.com/uiri/toml/blob/3c4cad681f8f4d3760c1cd0a538f8ccfd503fab3/toml/encoder.py#L158) should cover this case. Sequence types should be defaulted to the `list` dump function unless they explicitly have another dump function specified (e.g. strings are also sequence types).

Do you have a test case which demonstrates this not working as intended? A quick test shows that this works for all `tuple`s except for single element `tuple`s (which are unwrapped by default). This seems consistent with python's general treatment of `tuple`s.

jkarns275:
Yes, i do have a small example. This is more or less the exact issue I ran into. It may be the case that I'm using an old version of the library from pip.

```python
>>> toml.loads(toml.dumps({'a': (1, 2, 3)}))
  File "<stdin>", line 1, in <module>
  File "/home/wua/.local/lib/python3.6/site-packages/toml.py", line 373, in loads
    multibackslash)
  File "/home/wua/.local/lib/python3.6/site-packages/toml.py", line 478, in _load_line
    value, vtype = _load_value(pair[1], _dict, strictly_valid)
  File "/home/wua/.local/lib/python3.6/site-packages/toml.py", line 664, in _load_value
    v = int(v)
ValueError: invalid literal for int() with base 10: '(1, 2, 3)'
```

which works fine with lists:

```python
>>> toml.loads(toml.dumps({'a': [1, 2, 3]}))
{'a': [1, 2, 3]}
```

jkarns275:
Okay so it was just my outdated pip version - with the added line of code commented out it still works just as i'd expect it to. It does however appear to be the case that the most up-to-date version available on pip is not the most recent version on github.

PR #182, opened 2018-08-28
Update all pypi.python.org URLs to pypi.org

jdufresne:
For details on the new PyPI, see the blog post:

https://pythoninsider.blogspot.ca/2018/04/new-pypi-launched-legacy-pypi-shutting.html

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,7 +9,7 @@ TOML
     :target: https://travis-ci.org/uiri/toml
 
 .. image:: https://img.shields.io/pypi/pyversions/toml.svg
-    :target: https://pypi.python.org/pypi/toml
+    :target: https://pypi.org/project/toml/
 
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -24,7 +24,7 @@ See also:
 Installation
 ============
 
-To install the latest release on `PyPi <https://pypi.python.org/pypi/toml/0.9.2>`_,
+To install the latest release on `PyPi <https://pypi.org/project/toml/0.9.4/>`_,
```
uiri: Hm. Is there a reason this should really be linking to a specific version at all?

jdufresne:
Rebase. Now green.

> Hm. Is there a reason this should really be linking to a specific version at all?

Nope, I don't think so. I've removed it in the latest revision.

PR #186, opened 2018-08-29
Pass python_requires argument to setuptools

jdufresne:
Helps pip decide what version of the library to install.

https://packaging.python.org/tutorials/distributing-packages/#python-requires

> If your project only runs on certain Python versions, setting the python_requires argument to the appropriate PEP 440 version specifier string will prevent pip from installing the project on other Python
> versions.

https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords

> python_requires
>
> A string corresponding to a version specifier (as defined in PEP 440) for the Python version, used to specify the Requires-Python defined in PEP 345.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,7 @@ setup(
     packages=['toml'],
     license="MIT",
     long_description=readme_string,
+    python_requires=">=2.6, !=3.0.*, !=3.1.*, !=3.2.*",
     classifiers=[
         'Development Status :: 5 - Production/Stable',
         'Intended Audience :: Developers',


jdufresne:
Hi @uiri

I understand the hesitation. The decision is ultimately yours.

Using [pypinfo](https://github.com/ofek/pypinfo), we can check the statistics of downloads from PyPI from the last 30 days:

```bash
$ pypinfo --percent --markdown toml pyversion
```

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 3.6            |  44.90% |        172,023 |
| 2.7            |  37.20% |        142,531 |
| 3.5            |   7.36% |         28,216 |
| 3.7            |   5.81% |         22,273 |
| 3.4            |   4.62% |         17,716 |
| 3.8            |   0.05% |            201 |
| 3.3            |   0.03% |            126 |
| 2.6            |   0.02% |             71 |

As you can see, Python 2.6 (and 3.3) numbers are very low. This is true for the [larger Python community as well](https://medium.com/@hugovk/python-version-share-over-time-cf4498822650).

Further, Python 2.6 is end of life. It is no longer receiving bug fixes, including for security issues. Python 2.6 went EOL on 2013-10-29 -- almost five years ago. For additional details on support Python versions, see:

https://devguide.python.org/#status-of-python-branches

Removing support for EOL Pythons will reduce testing and maintenance resources.

Here are some additional perspectives on why projects should drop Python 2.6 support from Python core developers:

- https://snarky.ca/stop-using-python-2-6/
- http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html

Here is an interesting overview of showing top Python projects and which ones have dropped Python 2.6 support, including [pip](https://pip.pypa.io/en/stable/news/):

https://hugovk.github.io/drop-python/2.6/

Projects that require 2.6 support can continue to use an old version of toml by pinning the dependency in their `requirements.txt` or `setup.py`.

Again, the decision is ultimately up to you and I'm fine either way, but I thought I'd share my reasons why I think it is a good idea.

hugovk:
If 2.6 is dropped, remember to update the classifiers.

jdufresne:
Thanks for catching that. I've updated the trove classifiers.

uiri:
Hi Jon,

Thank you for digging up those statistics and those resources.

Python 2.6 was removed from the Travis configuration last November. I read many resources in favour of dropping 2.6 support at that time. These resources lead me to the conclusion that continuing to run unit tests on 2.6 was not a good idea.

I see no reason why the library won't continue to be compatible with Python 2.6 (and, for that matter, with Python 3.3) so I think the classifiers should stay. In fact, running the tests with 2.6 does still work. If there is a compatibility issue, then, of course, I will need to address it and then remove support for the following release.

Could you please add the Python 2.6 and Python 3.3 classifiers back?

Thanks,
-Will

jdufresne:
> Could you please add the Python 2.6 and Python 3.3 classifiers back?

No problem. I've done this in the latest revision. Thanks!

setup.py
```diff
@@ -18,6 +18,7 @@
     packages=['toml'],
     license="MIT",
     long_description=readme_string,
+    python_requires=">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.2.*",
```
hugovk: 3.2 is repeated here.

jdufresne: Oops! Thanks for catching. Fixed in the latest revision.

uiri:
Merged. Thanks again for your PR!

PR #190, opened 2018-09-06
Use isinstance, as object type may inherit from type

sivel:
I am working on integrating this project with another. In this project, we have some types that inherit from base types.

Using `self.dump_funcs.get(type(v))` results in `None` because the type of the object is defined as:

```
class AnsibleUnicode(AnsibleBaseYAMLObject, text_type):
    ''' sub class for unicode objects '''
    pass
```

Using `isinstance` would allow for this to work.

Before this change:

```toml
[all.vars]
group_var1 = value2
```

After this change:

```toml
[all.vars]
group_var1 = "value2"
```

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -160,3 +160,11 @@ def test_warnings():
 def test_commutativity():
     o = toml.loads(toml.dumps(TEST_DICT))
     assert o == toml.loads(toml.dumps(o))
+
+
+def test_bug_190():
+    class String(str):
+        pass
+
+    o = toml.dumps({'test': String('test string')})
+    assert o == 'test = "test string"\n'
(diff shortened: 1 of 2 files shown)


uiri:
Hi Matt,

The recommended solution to have support for encoding a custom type is to use a custom subclass of TomlEncoder whose `self.dump_funcs` includes an entry for the custom type.

Does this solution not work for your use case? Would your PR potentially negatively impact other users who use this pattern?

sivel:
Subclassing `TomlEncoder` will work. 

What is the estimated date for releasing 0.10.0?

This approach works in 0.10.0, but of course doesn't work in 0.9.4.

sivel:
Closing this.

Not knowing when 0.10.0 will release, for my needs I've created a recursive type conversion function to use before passing data in.  TOMLEncoder will be preferred where available, falling back to a type cleanser on older versions of this lib.

PR #195, opened 2018-09-12
Include tests in source distribution; clean up MANIFEST.in

### LINKED ISSUE #192: No tests in pypi source
Hi!

I'm currently trying to package toml for Arch Linux [community].
However, the tests are not in the source tarball on pypi (and I would like to run them). I can't take the github source tarball of 0.9.4 either (because there's none).

Can you include the tests in the next version and tag 0.9.4 in this repository, so that a tarball is generated here as well (I could use that meanwhile)?
Thanks!

jdufresne:
Don't need to specify `setup.py` as it is included automatically by `setuptools`.

Don't need to include package Python files (`toml/*`) as they're included automatically by `setuptools`.

Alphabetized list.

Fixes #192

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,8 +1,5 @@
-include toml/__init__.py
-include toml/decoder.py
-include toml/encoder.py
-include toml/tz.py
-include setup.py
-include toml.pyi
-include README.rst
 include LICENSE
+include README.rst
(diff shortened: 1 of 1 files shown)


uiri:
Oops looks like I missed this PR.

Thanks again Jon! Merged.

PR #197, opened 2018-09-16
Add support for decimal.Decimal encoding

thombashi:
In `toml 0.9.6`, `decimal.Decimal` values are converted into strings as follows:

```py
import toml
from decimal import Decimal

d = {"a": Decimal("0.1")}
print(toml.dumps(d))
```

```
a = "Decimal('0.1')"
```

It would be helpful if `toml.dumps` method output `float` values when input value types are `decimal.Decimal`:

```
a = 0.1
```

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -3,6 +3,7 @@ import copy
 import pytest
 import os
 import sys
+from decimal import Decimal
 
 from toml.decoder import InlineTableDict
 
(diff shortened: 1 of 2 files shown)


uiri:
Hi Tsuyoshi,

Thank you for your pull request! Sorry for the delay in following up on this.

I've merged this into master. This will go out in the next `0.10.X` release but it will *not* be part of a future `0.9.X` release (if any). I just want to clarify since you explicitly mentioned version 0.9.6.

PR #211, opened 2018-10-05
Prettify test suite and make it more robust

lig:
* Added pytest fixtures for inline test data and test samples from
`toml-test` repo.
* Added `https://github.com/BurntSushi/toml-test` as a git submodule under
`tests/` dir.
* Added `py37` to `tox.ini`.
* Changed `pypy` to `pypy27, pypy35` in `tox.ini`.
* Simplified Travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
  - "3.5"
  - "3.6"
  - "pypy"
+ - "pypy3.5"
 
 matrix:
   include:
@@ -15,19 +16,12 @@ matrix:
 
 notifications:
   email: false
-addons:
-  apt:
-    packages:
-    - git
-    - golang
 before_install:
  - export GOPATH=~/go
  - go get github.com/BurntSushi/toml-test
- - git clone https://github.com/BurntSushi/toml-test
 install:
  - pip install tox-travis
  - python setup.py install
- - chmod +x ./tests/*.sh
 script:
  - ~/go/bin/toml-test ./tests/decoding_test.sh
  - tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
  - "3.5"
  - "3.6"
  - "pypy"
- - "pypy3.5"
+ - "pypy3"
 
 matrix:
   include:
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tests/toml-test"]
+	path = tests/toml-test
+	url = https://github.com/BurntSushi/toml-test
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,23 @@
+import os
+
+import pytest
+
+
+@pytest.fixture(scope='session')
+def tests_dir():
+    return os.path.abspath(os.path.dirname(__file__))
+
+
+@pytest.fixture(scope='session')
+def toml_test_dir(tests_dir):
+    return os.path.join(tests_dir, 'toml-test', 'tests')
+
+
+@pytest.fixture(scope='session')
+def valid_dir(toml_test_dir):
+    return os.path.join(toml_test_dir, 'valid')
+
+
+@pytest.fixture(scope='session')
+def invalid_dir(toml_test_dir):
+    return os.path.join(toml_test_dir, 'invalid')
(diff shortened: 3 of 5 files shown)


lig:
Still struggling with codecov on this one.

uiri:
Hi Serge,

I am a bit apprehensive about changing to Pipfile if that means that toml now has a dependency on itself. `pipenv run tox` is kind of cool though. Could you please update the repository documentation too?

Is there a particular reason to change `TEST_STR` and `TEST_DICT` to methods beyond a stylistic choice? I am not opposed to it in principle; just curious about the change.

I am not really a fan of submodules, so I would prefer to keep the `toml-test` library separate.

I am glad to see that pypy3 is working again. It was removed a long time ago because it was inexplicably failing.

Thanks,
-Will

lig:
I think I need to explain my efforts here. The idea is to make the library as easy as possible to contribute to. IMO, this implies making the development process as intuitive as possible.

So, being able to checkout the source and dive in into the development right away is crucial in that sense.

> I am a bit apprehensive about changing to Pipfile if that means that toml now has a dependency on itself. pipenv run tox is kind of cool though.

Pipenv is not a dependency of the library here. It doesn't need to be a part of `setup.py`. It just allows to work on the library development having exactly the same environment for everybody involved. And it gives a place to support a list of development requirements apart from `setup.py`.

> Could you please update the repository documentation too?

Sure, thanks for pointing this out.

> Is there a particular reason to change TEST_STR and TEST_DICT to methods beyond a stylistic choice? I am not opposed to it in principle; just curious about the change.

This change is about `pytest` best practices. It is handy to have everything you use in tests wrapped in fixtures to declare explicitly which scope it belongs to and be able to easy reuse those fixtures in other test modules if needed, i.e. moving fixtures to `conftest.py` making them available for the entire test suite changing in the other tests' code.

> I am not really a fan of submodules, so I would prefer to keep the toml-test library separate.

Honestly, I'm neither a fan of git submodules. However, in this case, it's worth it as it does two things. First, it explicitly declares where to find this lib and eliminates the need to search for this info in the docs. Second, it couples the configuration of the tests suite and the submodule path eliminating the need to configure anything by hands preventing a possible contributor making mistakes on that stage.

> I am glad to see that pypy3 is working again. It was removed a long time ago because it was inexplicably failing.

Didn't know that. Just've tried and it worked:)

gaborbernat:
@lig as long as you have tox already setup diving in should be already as easy as ``python -m tox -e py36 --develop``. This will generate the py36 virtualenv correctly (and isolated) insite ``.tox/py36``. Moreover, you can replace to exactly which python you want (it also allows you to debug built failures for a given Python version). Given installing tox is just ``python -m pip install tox --user`` not sure how now requiring pipenv to be installed makes things any easier.

lig:
@gaborbernat this makes sense. However, in this way, it is harder to support development requirements and work with them from the IDE. Also, it doesn't harden packages' versions in the environment making different development environments nonidempotent.

On the other hand, I'd say that integrating Pipenv is not a target of this MR and could be postponed and discussed separately.

gaborbernat:
Why it's harder? If anything you have more target than dev and install requirements. This way one can better separate requirements for different tasks (e.g. test, documentation builder, formatter, static analysis and etc.). Why would it be harder to work from within IDE? VsCode automatically can discover tox environments (it automatically scans for virtual envs under the project folder), and PyCharm has https://github.com/nokia/PyVenvManage that allows you to right click and set a given virtual environment. 

It will harden package versions at the same level the user is hardened by it (e.g. depending on how hard you harden it with PEP-440 strings). Arguably, for library development, this is better, because it means you'll test with the same version the user would get if they would do a pip install (meaning between builds the versions may change, but this reflects how users may get different versions depending on when they install the package).

lig:
`Coverage not affected when comparing 3a694e9...62a5bf8`

lig:
@uiri @gaborbernat this one is ready for review. I've removed pipenv stuff

.travis.yml
```diff
@@ -6,6 +6,7 @@ python:
  - "3.5"
  - "3.6"
  - "pypy"
+ - "pypy3.5"
```
hugovk: This now points to a more recent version.
```suggestion
 - "pypy3"
```

PR #225, opened 2019-01-01
Fix ResourceWarning: unclosed file in test_api.py

BoboTiG:
Hello,

This is a little sanitization fix to prevent resource leaks.


uiri:
Thank you so much for the PR! Merged.

Issue #223, opened 2018-12-06
dump


jwaggy:
ahoy,

Long time enjoyer of this library.

Is there a reason that toml.dump doesn't open a file itself like toml.load does? If so could someone or myself add this?

uiri:
Hello!

The intent of `dump`/`load` and `dumps`/`loads` was to mimic python's builtin `json` module.

I suppose there is no technical reason that `dump` couldn't support that usage in a backwards compatible way. If you want to open a pull request which adds that (plus tests :smile: ), I would be more than happy to take a look at it!

PR #226, opened 2019-01-26
Add support for output to a file to toml.dump function: #223

thombashi:
This PR is to enhance `toml.dump` that output `toml` to a file. And still, keep the backward compatibility (Issue #223).

If passing a string to `toml.dump` function, open it as a file path and write `toml` content to the file.
And behave the same as before, if passing a file descriptor to `toml.dump` function.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -183,7 +183,7 @@ class FakeFile(object):
         return self.written
 
 
-def test_dump():
+def test_dump(tmpdir):
     f = FakeFile()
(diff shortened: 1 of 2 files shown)


uiri:
Hi Tsuyoshi,

This PR seems like it might be OK, but may I ask how the approach in this PR compares with having the caller wrap their `toml.dump` call like this instead?

```python
with io.open(filename, "w", encoding="utf-8") as f:
    toml.dump(o, fd)
```

Issue #234, opened 2019-02-25
Calling load(list_of_files, _dict=OrderedDict) does not return an OrderedDict


erkyrath:
The _dict argument is ignored for the top-level object.

LexiconCode:
Thank you I was going to look into this as well.

uiri:
Thank you for reporting this issue!

It looks like I forgot to pass `_dict` to the constructor for `TomlDecoder` in `load`. This has been fixed in the latest master branch.

PR #235, opened 2019-02-25
When calling load() with an array of filenames, use the requested dictionary type.

erkyrath:
Fixes https://github.com/uiri/toml/issues/234 .


uiri:
Thanks for the PR!

PR #241, opened 2019-04-06
WIP: Add comment support

### LINKED ISSUE #77: Comments support

PrajwalM2212:
fixes #77

### DIFF
--- a/toml/decoder.py
+++ b/toml/decoder.py
@@ -96,6 +96,8 @@ def load(f, _dict=dict, decoder=None):
         f: Path to the file to open, array of files to read into single dict
            or a file descriptor
         _dict: (optional) Specifies the class of the returned toml dictionary
+        comments: (optional) Specifies whether comment information has to
+                  preserved or not
 
     Returns:
         Parsed toml file represented as a dictionary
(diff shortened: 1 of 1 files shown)


uiri:
Hi Prajwal,

Thank you for your pull request! But this should really be part of a decoder. One of the reasons I adopted the decoder/encoder model was to avoid an explosion of parameters to `dump` / `load` and `dumps` / `loads`.

Also, I don't think your test is quite right. The point of saving comments is not to add them to the dictionary as key-value pairs to the dictionary that `loads` returns but to make sure that they roundtrip through loads and dumps without modification.

tests/test_api.py
```diff
@@ -6,14 +6,21 @@
 from decimal import Decimal
 
 from toml.decoder import InlineTableDict
-
+from toml.decoder import loads
```
uiri: nit: this import is unnecessary. `toml` is imported above, so you can use `toml.loads`

uiri: nit: this import is unnecessary. `toml` is imported above, so you can use `toml.loads`

toml/decoder.py
```diff
@@ -43,7 +43,6 @@ def _getpath(p):
 except NameError:
     FNFError = IOError
 
-
```
uiri: nit: whitespace change

PrajwalM2212:
@uiri Yes, I totally missed the encoder part. 
So would this approach be okay ?
1.  load comments into a dictionary like I have done here by a key like 'comment_0', 'comment_1' etc.
2. dump by converting these 'comment_0', 'comment_1' into `# comment `

PR #242, opened 2019-04-13
Comment support

### LINKED ISSUE #77: Comments support

PrajwalM2212:
closes #77


PrajwalM2212:
@uiri 
`  dump(load('config.toml', OrderedDict), f) `

config.toml
```toml
[flakes]
# Do not set default_action to ApplyPatchAction as it may lead to some
# required imports being removed that might result in coala behaving weirdly.
default_actions = '*: ShowPatchAction' # Hello world
bears = 'PyUnusedCodeBear'
language = 'Python'
remove_all_unused_imports = true

[[products]]
name = "Nail"
sku = 284758393
# dgda
color = "gray" # Hello
```

output
```toml
[[products]]
name = "Nail"
sku = 284758393
# dgda
color = "gray" # Hello

[flakes]
# Do not set default_action to ApplyPatchAction as it may lead to some
# required imports being removed that might result in coala behaving weirdly.
default_actions = "*: ShowPatchAction" # Hello world
bears = "PyUnusedCodeBear"
language = "Python"
remove_all_unused_imports = true
```

PrajwalM2212:
Works for both inline and normal comments.
It associates each comment with the immediate keyname preceding the comment.  This allows the comments to be dumped at the right place.

uiri:
Hi Prajwal,

Thanks for the PR! Could you split this into a custom encoder and custom decoder?  
Maybe the standard decoder could have a `handle_comment` method that the custom decoder could override?  
I think the encoder can be customized for preserving comments by overriding existing dumping logic.

PrajwalM2212:
@uiri  Is this the way you wanted this?
If it is ok, then I will begin polishing this and add tests.

uiri:
@PrajwalM2212 Yes, this is the way I wanted it. Thanks!

I need to take a deeper look before I merge it, but I will wait for you to polish it and add tests before adding PR comments.

PrajwalM2212:
@uiri I have improved the comment mechanism.
It now associates each comment with the immediate `keyname` or `table` or `array of tables` preceding the comment. This allows the comments to be dumped at the right place and removes bugs from the previous approach.

PrajwalM2212:
@uiri  Can you review this?

PrajwalM2212:
@uiri Can you review and merge this when you are free? I need this for my Google Summer Of Code project.

toml/decoder.py
```diff
@@ -957,3 +964,54 @@ def load_array(self, a):
...
+                sl[j] = ' '
+                j += 1
+        except IndexError:
+            raise IndexError()
+        return beginline
+
+
+class TomlPreserveCommentDecoder(TomlDecoder):
+
+    def __init__(self, _dict=OrderedDict):
```
uiri: Default `_dict` should not be `OrderedDict`

tests/test_api.py
```diff
@@ -7,7 +7,6 @@
 
 from toml.decoder import InlineTableDict
 
-
```
uiri: nit: whitespace change

toml/decoder.py
```diff
@@ -43,7 +44,6 @@ def _getpath(p):
 except NameError:
     FNFError = IOError
 
-
```
uiri: nit: whitespace change

toml/decoder.py
```diff
@@ -106,7 +107,6 @@ def load(f, _dict=dict, decoder=None):
         IOError / FileNotFoundError -- When an array with no valid (existing)
         (Python 2 / Python 3)          file paths is passed
     """
-
```
uiri: nit: whitespace change

toml/decoder.py
```diff
@@ -277,11 +277,8 @@ def loads(s, _dict=dict, decoder=None):
                 openstrchar = ""
         if item == '#' and (not openstring and not keygroup and
                             not arrayoftables):
-            j = i
             try:
-                while sl[j] != '\n':
-                    sl[j] = ' '
-                    j += 1
+                beginline = decoder.comment_lint(sl, i, beginline)
```
uiri: Why can't you use handle comment here?

toml/encoder.py
```diff
@@ -41,7 +42,6 @@ def dumps(o, encoder=None):
     Returns:
         String containing the toml corresponding to dict
     """
-
```
uiri: nit: whitespace

toml/decoder.py
```diff
@@ -957,3 +964,54 @@ def load_array(self, a):
...
                 retval.append(nval)
         return retval
+
+    def handle_inline_comment(self, line, currentlevel):
+        return line
+
+    def handle_comment(self, currentlevel, comment_no, recent, line):
+        return comment_no
+
+    def comment_lint(self, sl, i, beginline):
```
uiri: Why is beginline a necessary parameter and why does this method return it?

toml/decoder.py
```diff
@@ -957,3 +964,54 @@ def load_array(self, a):
...
+        k = pos
+        comment = ""
+        try:
+            while line[k] != '\n':
+                comment += line[k]
+                k += 1
+        except IndexError:
+            pass
+        pair = line.split('=')
+        currentlevel[pair[0].strip() + '-' + 'comment' + '_inline'] = comment
```
uiri: Doesn't this interfere with keys that end in `-comment_inline` ?
Not likely, but prone to weirdness. I think a custom dictionary type that keeps the comments outside of the normal key-space of the dictionary would eliminate that class of bugs.

toml/decoder.py
```diff
@@ -957,3 +964,54 @@ def load_array(self, a):
...
+        line = line.split('#')[0]
+        return line
+
+    def handle_comment(self, currentlevel, comment_no, recent, line):
+        if recent + '-comment' in currentlevel:
+            currentlevel[recent + '-' + str(comment_no) + '-' + 'comment'] \
+                = line
+            comment_no += 1
+        else:
+            currentlevel[recent + '-' + 'comment'] = line
```
uiri: Same thing as above but for keys ending in `-comment` which seems likely to me to be far more common...

toml/decoder.py
```diff
@@ -957,3 +964,54 @@ def load_array(self, a):
...
+            comment_no = 0
+        return comment_no
+
+    def comment_lint(self, sl, i, beginline):
+        beginline = False
+        return beginline
```
uiri: Equivalent to `return False`

toml/decoder.py
PrajwalM2212: This part of the code just validates the toml file character by character and does not load anything to the retval(current_level) dictionary.

toml/decoder.py
PrajwalM2212: This part of the code removes the comment data if TomlPreseveCommentDecoder is not used. beginline may not make sense if TomlPreseveCommentDecoder is not used. 

But if TomlPreseveCommentDecoder is used, the comment data has to be preserved and beginline must be set to false for the decoder to not misinterpret the comment.

toml/decoder.py
PrajwalM2212:
```python
class CommentKey:
    def __init__(self, key, val):
        self.key = key
        self.val = val

    def __hash__(self):
        return hash(str(self))

    def __str__(self):
        return '{}, {}'.format(self.key, self.val)

    def __eq__(self, other):
        return str(self) == other
```
Something like this ?

toml/decoder.py
PrajwalM2212: Using ordered dict makes for faster dumping and also easier to code. 
I will create make provision when normal dict is used.

PrajwalM2212: Resolved

PrajwalM2212:
@uiri All the issues have been solved.

PrajwalM2212:
@uiri   All bases have been covered inline, full line, multiline array). Can we wind this up?

tests/test_api.py
```diff
@@ -211,3 +211,22 @@ def test_warnings():
...
+    test_str = """[[products]]
+name = "Nail"
+sku = 284758393
+num = [ 1, #Number 1
+        2, # Number 2
+        3,
+      ]
+# This is a comment
+color = "gray" # Hello World
+"""
```
uiri: I think it would be good to test scenarios where comments are valid toml.

e.g. a commented out key with inline table, commented out key with multiline string, commented out key with array

PrajwalM2212:
@uiri Can u review it now?
I have used line number logic. I have also added more complex test cases. 

Took a lot of effort and time to complete this. 💪

toml/decoder.py
```diff
@@ -177,6 +193,11 @@ def loads(s, _dict=dict, decoder=None):
     keygroup = False
     dottedkey = False
     keyname = 0
+
+    # for collecting comment info
+    comment_info = list()
+    comment_lines = []
```
uiri: I think it would be better if these variables lived inside the decoder.

toml/decoder.py
```diff
@@ -326,9 +349,50 @@ def loads(s, _dict=dict, decoder=None):
     multilinestr = ""
     multibackslash = False
     pos = 0
+    comment_norm_factor = 0
     for idx, line in enumerate(s):
         if idx > 0:
             pos += len(s[idx - 1]) + 1
+
+        if len(comment_info) > 0 and comment_info[0][0] - 1 == idx:
```
uiri: I think that most of this new logic belongs in your preserve comment decoder.

Most likely, it is better to pass idx to the decoder and let the decoder determine if any of its `comment_info` have a match.

Maybe rather than `comment_info.append((a, b)); comment_lines.append(c)`, `comment_info[a].append((b, c)) if a in comment_info else comment_info[a] = [(b, c)]`

`a` needs to already be adjusted by the normalization factor for this to work though... the decoder could certainly keep track of that, I think?

toml/decoder.py
```diff
@@ -326,9 +349,50 @@ def loads(s, _dict=dict, decoder=None):
...
+            # match the comment lines with their original positions
+            # in the file and handle them appropriately
+            original_line_no = comment_lines[0] - 1
+            original_line = (original.split('\n')[original_line_no])
```
uiri: It might be useful to split `original_lines` on `\n` once and save it rather than once or more for each comment.

toml/decoder.py
```diff
@@ -326,9 +349,50 @@ def loads(s, _dict=dict, decoder=None):
...
+            recent = 'start'
```
uiri: Was there something wrong with the previous approach of keeping track of recent as the file was parsed line-by-line?

toml/decoder.py
```diff
@@ -326,9 +349,50 @@ def loads(s, _dict=dict, decoder=None):
...
+                        recent_line_no -= 1
+                        continue
+
+                    recent = prev_line.split('=')[0].strip()
+                    break
+                decoder.handle_comment(currentlevel, original_line,
+                                       'normal', original,
+                                       recent)
+
+            elif len(s[idx].split('=')) == 2:
```
uiri: This won't capture this line of TOML:

```
x = "Testing 3 != 4" # some string with = in it followed by a comment
```

toml/decoder.py
```diff
@@ -957,3 +1021,114 @@ def load_array(self, a):
...
+    def handle_comment(self, currentlevel, line, comment_type,
+                       original, recent=None):
+        if comment_type == 'normal':
+            currentlevel[CommentKey(recent + '-' + 'comment', line)] = line
+        elif comment_type == 'inline':
+            self.handle_inline_comment(line, currentlevel)
+        else:
+            self.handle_array_comment(line, currentlevel, original)
+
+    @staticmethod
```
uiri: I'm not sure I understand this decorator.
It says static method, but you are calling this as if it is an instance method.
Would it not be easier to understand if you called it as `handle_array_comment` (not `self.handle_array_comment`) and dropped this decorator?

toml/decoder.py
```diff
@@ -284,6 +305,8 @@ def loads(s, _dict=dict, decoder=None):
                     j += 1
             except IndexError:
                 break
+            # store the comment info
+            decoder.store_comment_info(i, comment_info, comment_lines, original)
```
uiri: I think this needs to go _before_ the while loop with sl[j] = ' ', or else you lose the comment text...

toml/decoder.py
PrajwalM2212: It captures as `s[idx]` the line from with comment is removed and not original.

PrajwalM2212:
@uiri  All the requested changes have been made. 
It now looks nice as all the comment logic is inside the comment preserve decoder.

`All the logic in the decoder is accomplished in one pass.`

But encoder can't be done in one pass. 
If the decoder was limited to OrderedDict, then encoder could be done in one pass as well. 

I don't think anybody who wants to preserve comments would use a normal dict. Alternatively, there could be two methods - one for normal dict and ordereddict. In documentation, it could be added that using ordereddict would be faster. 

So what should be done for the encoder to speed it up,
1. limit decoder to ordereddict 
or 
2. add another fast encoder method for ordereddict

toml/decoder.py
PrajwalM2212: This was a nice edge case.

PrajwalM2212:
@uiri I found the current method too buggy and incompetent especially when dealing with multiline arrays. 

Hence I have been working on a new logic that is bug free and multiline comment friendly.
Hope to  open a PR soon.

uiri:
Closing this as it appears to have been abandoned in favour of #249

PR #245, opened 2019-04-26
Remove precision from dump float

davebrent:
Precision in a python format string, without a type specifier, formats
the content as a _string_, where the precision indicates the maximum
field size. This results in numbers with a large integer part, but a
small decimal part, being truncated unnecessarily. eg.

    >>> '{0:.16}'.format(1555609276.1339364)
    '1555609276.133936'


uiri:
Thank you for your pull request! Merged.

PR #248, opened 2019-06-06
Small documentation and argument update

mileslucas:
This pull request does two things:

1. In `toml.dump` allows passing an encoder argument. Currently, to use a non-standard encoder users are forced to use `toml.dumps` followed by file-writing themselves. `toml.dump` should include the encoder argument to avoid this.

2. Updated README documentation to show the new arguments and gives an example of using the `TomlNumpyEncoder`. I fell into quite the rut in my code because my numpy arrays were saving as strings. The only way I figured out how to solve it was finding an old closed issue and then trying to figure out the usage through reading the source code. Hopefully adding this documentation will help users moving forward.

I have also incremented to version to `0.10.1` in hopes that a release (or pre-release) will be made on PyPI so I can take the github dependency out of my setup.py :)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "restructuredtext.confPath": ""
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "restructuredtext.confPath": ""
-}
--- a/README.rst
+++ b/README.rst
@@ -114,6 +114,22 @@ corresponding TOML-formatted data.
 
 For more functions, view the API Reference below.
 
+Note
+----
+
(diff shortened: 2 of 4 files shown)


uiri:
This is awesome! Thank you! Merged.

mileslucas:
Any chance on that pypi release? Or are there other aspects holding it back?

PrajwalM2212:
It would be nice if a new version is released after the comment support PR has been merged.

frrad:
Any chance of a pypi release? Would like to use the new `encoder` kwarg in `toml.dump` instead of working around.

mileslucas:
Bumping for PyPI release. If there are any other features waiting to be included I suggest they wait for a version `0.11.0`.

PR #249, opened 2019-06-12
Comment Support

PrajwalM2212:


PrajwalM2212:
```toml
# Test file for TOML
# Only this one tries to emulate a TOML file written by a user of the kind of parser writers probably hate
# This part you'll really hate

[the]
#dasc
test_string = "You'll hate me after this - #"          # " Annoying, isn't it?

    [the.hard]
    test_array = [ "] ", " # "]      # ] There you go, parse this!
    test_array2 = [ "Test #11 ]proved that", "Experiment #9 was a success" ]
    # You didn't think it'd as easy as chucking out the last #, did you?
    another_test_string = " Same thing, but with a string #"
    harder_test_string = " And when \"'s are in the string, along with # \""   # \"and comments are there too"
    # Things will get harder

        [the.hard."bit#"]
        "what?" = "You don't think some user won't do that?"
        multi_line_array = [
            "]",
            # ] Oh yes'd I did
            ]

# Each of the following keygroups/key value pairs should produce an error. Uncomment to them to test

#[error]   if you didn't catch this, your parser is broken
#string = "Anything other than tabs, spaces and newline after a keygroup or key value pair has ended should produce an error unless it is a comment"   like this
#array = [
#         "This might most likely happen in multiline arrays",
#         Like here,
#         "or here,
#         and here"
#         ]     End of array comment, forgot the #

a = 'animal' #cacd
#acdac
#qdqdfw

[[products]]
name = "Nail"
sku = 284758393
# This is a comment
color = "gray" # Hello World
# name = { first = 'Tom', last = 'Preston-Werner' }
# arr7 = [
#  1, 2, 3
# ]
# lines  = '''
# The first newline is
# trimmed in raw strings.
# All other whitespace
# is preserved.
# '''
x = "Testing 3 != 4" # some string with = in it followed by a comment
test_string = "You'll hate me after this - #" # " Annoying, isn't it?

[animals]
color = "gray" # col
fruits = "apple" # a = [1,2,3]
```

PrajwalM2212:
```toml
# Test file for TOML
# Only this one tries to emulate a TOML file written by a user of the kind of parser writers probably hate
# This part you'll really hate
[[products]]
name = "Nail"
sku = 284758393
# This is a comment
color = "gray" # Hello World
# name = { first = 'Tom', last = 'Preston-Werner' }
# arr7 = [
#  1, 2, 3
# ]
# lines  = '''
# The first newline is
# trimmed in raw strings.
# All other whitespace
# is preserved.
# '''
x = "Testing 3 != 4" # some string with = in it followed by a comment
test_string = "You'll hate me after this - #" # " Annoying, isn't it?

[the]
#dasc
test_string = "You'll hate me after this - #" # " Annoying, isn't it?

[animals]
color = "gray" # col
fruits = "apple" # a = [1,2,3]
arr4 = [ "all", "strings", "are the same", "type",]
arr5 = [ [ 1, 2,], [ "a", "b", "c",],] #arr5 = [ [ 1, 2,], [ "a", "b", "c",],]

[the.hard]
test_array = [ "] ", " # ",] # ] There you go, parse this!
test_array2 = [ "Test #11 ]proved that", "Experiment #9 was a success",]
# You didn't think it'd as easy as chucking out the last #, did you?
another_test_string = " Same thing, but with a string #"
harder_test_string = " And when \"'s are in the string, along with # \"" # \"and comments are there too"
# Things will get harder

[the.hard."bit#"]
"what?" = "You don't think some user won't do that?"
multi_line_array =[
            "]",
            # ] Oh yes'd I did                 
            ]
# Each of the following keygroups/key value pairs should produce an error. Uncomment to them to test
#[error]   if you didn't catch this, your parser is broken
#string = "Anything other than tabs, spaces and newline after a keygroup or key value pair has ended should produce an error unless it is a comment"   like this
#array = [
#         "This might most likely happen in multiline arrays",
#         Like here,
#         "or here,
#         and here"
#         ]     End of array comment, forgot the #
a = "animal" #cacd
#acdac
#qdqdfw
```

PrajwalM2212:
@uiri  I think this is the best approach. Also, I think there is no better approach than this. 
It is able to catch all types of comments.

toml/decoder.py
```diff
@@ -311,6 +363,7 @@ def loads(s, _dict=dict, decoder=None):
                     if sl[i - 3] == sl[i - 1]:
                         sl[i - 3] = ' '
             elif openarr:
+                arr += s[i]
```
uiri: Why preserve newlines inside of TOML arrays?

toml/decoder.py
```diff
@@ -177,11 +192,28 @@ def loads(s, _dict=dict, decoder=None):
     keygroup = False
     dottedkey = False
     keyname = 0
+
+    arr = ''
+    comm_arr = False
+    arr_beg = arr_end = None
+    comm_list = []
```
uiri: Can the `comm_list` go directly inside the decoder? This avoids the save step at the end of the first pass, and helps move a bit more logic inside the decoder.

PrajwalM2212:
@uiri  I have decided to split the PR into two. This PR will address the majority case logic. I thought I would write down the logic so that you could review it faster.

## Part 1
The base logic is,
1.  I collect the comments
2. I collect the original line number of the comment
3. The decoder compresses multiline arrays and multiline strings.  This is kept track by a variable `compress_size`.  I collect the line number where the comment will appear in `s` after this compression. 
All these three information is stored in `comm_list`

## Part 2
```python
for idx, line in enumerate(s):
        if idx > 0:
            pos += len(s[idx - 1]) + 1

        decoder.match_comments(idx, original, s, currentlevel, recent)
```
Here `decoder.match_comments` checks the idx aganist the line number stored in `comm_list` and 
calls `handle_comment`.

## Part 3
`handle_comment` stores these comments in the dictionary.

I think that this PR will be able to catch all inline and full life comments.

PrajwalM2212:
@uiri Can u review and merge this when you are free ?

toml/encoder.py
```diff
@@ -263,3 +280,50 @@ def __init__(self, _dict=dict, preserve=False):
...
+                            lst[i] = "blank"
+                            key = remove_quotes(key[0].
+                                                split('-comment_inline')[0])
+                            for indx, j in enumerate(lst):
+                                if len(j.split('=', 1)) == 2:
+                                    k = j.split('=')[0].strip()
+                                    if k == key:
+                                        val = unescape(val)
+                                        lst[indx] = j + ' ' + val
+                    elif '-comment' in key[0]:
```
uiri: Will this have trouble with the following TOML?

```
a = 3
b-comment = "a is 3"
```

PrajwalM2212: No

PrajwalM2212: I have added these statements to test cases as well.

toml/decoder.py
```diff
@@ -278,12 +304,18 @@ def loads(s, _dict=dict, decoder=None):
...
+            comment_line_no = original.count('\n', 0, i) + 1
+            comment = prev_key
             try:
                 while sl[j] != '\n':
+                    comment += s[j]
                     sl[j] = ' '
                     j += 1
             except IndexError:
                 break
+            if not openarr:
```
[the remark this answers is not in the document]
PrajwalM2212: For something like 
```toml
a = [ 1, #Hello
        2, #World
        3,
     ]
```
the decoder makes it as ```a = [1, 2, 3]```.
So the original line number or the compressed line number of the comments here doesn't
matter as multiline array becomes a single line array.

In fact, in order to preserve comments in multiline arrays, I have to store the entire multiline array with comments as it is in the comm-list.
Then while dumping I have to replace ```a = [1, 2, 3]``` with original array. 

Multiline arrays with comments will need special handling but will still use the same logic.

PrajwalM2212:
@uiri  All changes have been made.

toml/decoder.py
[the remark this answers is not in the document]
uiri: OK, that makes sense. So for now it preserves comments _except_ for those inside of multiline arrays (and ostensibly those inside of multiline strings but I am not sure how/if that would be valid..?)

toml/encoder.py
uiri: This whole section sort of makes me nervous for a couple reasons:

1. It is dependent on a specific pattern of newlines to delineate "sections"
2. You are still saving the comments beside regular keys. At least, that is what it looks like to me? Do you have an example retval input for this `write_comments` function?

uiri:
Hi Prajwal!

I want to thank you again for your pull requests. Reviewing them helped me a lot in terms of thinking about how to go about solving this problem. I sat down to fixup this PR so that I could merge it, and ended up rewriting most of your implementation. Oops.

Since I merged my implementation into master along with the test you included in this PR and it passes, I am closing this PR.

Can you let me know if you think you can adapt your multiline array handling code to work with my implementation's approach?

Thanks.

PrajwalM2212:
Your implementation is cleaner and concise. Glad my PR was of help.

PR #250, opened 2019-06-17
Extend TomlNumpyEncoder to numpy integer types.

filippo82:
I have extend TomlNumpyEncoder to handle numpy integer types. I've also added the unit test for these cases.

### DIFF
--- a/toml/encoder.py
+++ b/toml/encoder.py
@@ -118,10 +118,6 @@ def _dump_float(v):
     return "{}".format(v).replace("e+0", "e+").replace("e-0", "e-")
 
 
-def _dump_int(v):
-    return "{}".format(int(v))
-
-
 def _dump_time(v):
     utcoffset = v.utcoffset()
     if utcoffset is None:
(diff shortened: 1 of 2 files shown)


toml/encoder.py
```diff
@@ -118,6 +118,10 @@ def _dump_float(v):
     return "{}".format(v).replace("e+0", "e+").replace("e-0", "e-")
 
 
+def _dump_int(v):
```
uiri: Thank you for your pull request!

Could you move `_dump_int` to be inside of `TomlNumpyEncoder`?
I do not think there is any need to modify the existing encoder to extend the Numpy one to support integers.

filippo82: Hi Will,

just to clarify, you are suggesting to revert `int: _dump_int` to `int: lambda v: v` inside the `TomlEncoder` class and to move the `_dump_int` function inside the `TomlNumpyEncode` class?

Cheers.

uiri: Yes, that is my suggestion exactly :smiley:

filippo82: Done!

PR #266, opened 2019-11-12
Fix the issue where the parser fails on time attr with no space after =

salmangano:
like time=10:00:00


uiri:
Hi Sal,

Thanks for the pull request!

Merged.

PR #275, opened 2020-01-06
encoder: Dump nested tables with their parent

### LINKED ISSUE #88: Nested tables should be written below their parent table
Though not a requirement of the spec, it would be a very nice feature of this library to, when encoding nested dictionaries, write nested tables directly below their parent table, instead of first all top-level sections then the next level, then the next.

bochecha:
Currently, we first print all tables at the top level, then all nested
tables at the second level, etc…

As a result, this:

    {
        'table1': {
            'foo': 1,
            'nested': {
                'bar': 2,
            },
        },
        'table2': {
            'baz': 3,
        },
    }

… gets printed as:

    [table1]
    foo = 1

    [table2]
    baz = 3

    [table1.nested]
    bar = 2

With this commit, we print the nested tables with their parent, so the
same example from above will instead become:

    [table1]
    foo = 1

    [table1.nested]
    bar = 2

    [table2]
    baz = 3

Fixes #88


bochecha:
In #88, @sayanarijit suggests doing it [with a new ordering option](https://github.com/uiri/toml/issues/88#issuecomment-455852620)

It wasn't clear from that issue whether anything had been decided and that was the way to go, so I figured I wouldn't spend too much time on this before reaching an agreement on this.

As a result, take this pull request as a strawman for how the desired result can be achieved code-wise, and to restart the conversation on what the API should be. :slightly_smiling_face:

PR #279, opened 2020-01-07
dotted keys with multiline strings

### LINKED ISSUE #265: Broken parsing of triple-quoted strings with dotted keys
I found another problem with triple quoted strings that seems not to be covered in related open bugs (#123, #230, #237).
```
>>> print(toml.__version__)
0.10.0
>>> test = 'a.b = """123\n"""'
>>> print(test)
a.b = """123
"""
>>> toml.loads(test)
{'a': {}, 'b': '123\n'}
```
but I would expect `{'a': {'b': '123\n'}}`

Note that it is the newline inside the triple-quoted string that triggers this error:
```
>>> test2 = 'a.b = """123\"""'
>>> toml.loads(test2)
{'a': {'b': '123'}}
```

samuelcolvin:
I've added a test for this: https://github.com/BurntSushi/toml-test/pull/57

fix #265

The problem is that when you call `currentlevel[pair[0]] = value` on line 786, you've re-assigned currentlevel thus isn't not the same as where you were calling `currentlevel[multikey] = value` on line 395, this get's around that problem by returning a new `multilevel` object.

### DIFF
--- a/toml/decoder.py
+++ b/toml/decoder.py
@@ -360,6 +360,7 @@ def loads(s, _dict=dict, decoder=None):
                               " Reached end of file.", original, len(s))
     s = ''.join(sl)
     s = s.split('\n')
+    multilevel = None
     multikey = None
     multilinestr = ""
     multibackslash = False
@@ -392,7 +393,8 @@ def loads(s, _dict=dict, decoder=None):
                     value, vtype = decoder.load_value(multilinestr)
                 except ValueError as err:
                     raise TomlDecodeError(str(err), original, pos)
-                currentlevel[multikey] = value
+                multilevel[multikey] = value
+                multilevel = None
                 multikey = None
                 multilinestr = ""
             else:
@@ -509,7 +511,7 @@ def loads(s, _dict=dict, decoder=None):
             except ValueError as err:
                 raise TomlDecodeError(str(err), original, pos)
             if ret is not None:
-                multikey, multilinestr, multibackslash = ret
+                multilevel, multikey, multilinestr, multibackslash = ret
     return retval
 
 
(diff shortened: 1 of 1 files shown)


puhitaku:
Thanks for the patch as I also suffered this problem.
I really want it to be merged soon... Could you consider merging this? @uiri

samuelcolvin:
I think this library is no longer properly maintained (although I know it's very widely used).

you might try [rtoml](https://github.com/samuelcolvin/rtoml) which should fix this problem and many more (admission: I built rtoml).

puhitaku:
That sounds great, wanna try this soon!
For a variety of TOML parser let me introduce that [Poetry uses tomlkit](https://github.com/python-poetry/poetry/blob/master/pyproject.toml#L38) for parsing TOML.

samuelcolvin:
https://github.com/samuelcolvin/rtoml/tree/master/benchmarks

Perhaps poetry likes life in the slow lane :-)

gaborbernat:
@samuelcolvin does ``rtoml`` supports roundtrip guarantees e.g. ``dump(load(toml_text)) == toml_text``?

samuelcolvin:
> @samuelcolvin does `rtoml` supports roundtrip guarantees e.g. `dump(load(toml_text)) == toml_text`?

I don't think so, it'll be the same as [toml-rs](https://github.com/alexcrichton/toml-rs) which it's based on.

gaborbernat:
In that case your comparison is not fair, and you should probably remove it from the table. That feature is very important for poetry, and tools in general (that need to alter config files without modifying the formatting of the file). They're designed and optimized for different use case.

samuelcolvin:
I've just tried and rtoml definitely doesn't provide roundtrip guarantees. I don't know if that's a problem in the python or a characteristic of toml-rs (you would have though it would be necessary to load and write to `Cargo.toml`?).

I see the distinction, though for many applications it's not a problem. Happy to accept a PR to add a clarification to benchmarks.

By the way, I think (if I remember correctly) tomlkit also fails some of the standard toml tests. I can give details if necessary.

PR #291, opened 2020-04-22
Raise an exception if EOF is reached and a string is unterminated

mraspberry:
Given a TOML file with an unterminated string value as the last option in the file
```toml
opt='''
this is a test
of an unfinished string
```
Current toml version (0.10.0) will simply not load that option instead of raising an Exception. The following test code returns an empty dict given the above toml file
```python
import toml
import pathlib

filepath = pathlib.Path('test.toml')
data = toml.loads(filepath.read_text())
assert bool(data)
```

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -270,13 +270,3 @@ b-comment = "a is 3"
                    encoder=toml.TomlPreserveCommentEncoder())
 
     assert len(s) == len(test_str) and sorted(test_str) == sorted(s)
-
-
-@pytest.mark.parametrize('test_val', [
-    'opt = "unterminated double\n',
-    "opt = 'unterminated single\n",
-    "opt = '''\nunterminated\nraw multiline\n",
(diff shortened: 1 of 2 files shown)


mraspberry:
The checks all show as passed if you look at the Details but it's not reflected here

uiri:
Hi!

I think the test belongs in the toml-test suite since it tests invalid toml. The included tests are intended for testing the python API as exposed by the library.

The fix itself looks good to me.

mraspberry:
Ok I'll create the test in the toml test suite instead. Thanks for the feedback.

mraspberry:
I added an inline ignore for flake8 E741 (ambiguous variable name) on line 147 of toml/decoder.py since i didn't modify the code in question and i'm not sure why it's complaining about it now and not in the other commits I've made. I can actually fix the error if you prefer, just let me know.

uiri:
Thanks! Merged :)

PR #293, opened 2020-05-05
Fix #204 triple quote detection incorrect on non stipped line

### LINKED ISSUE #204: Parsing fails on multiline string with comments
I can't put my finger on this exactly, but it seems that the parser gets confused by certain combinations of multiline strings with comments.

For example:

    toml.loads('a = """#\n"""#')

returns `{}`
while it should return `{"a": "#\n"}`

May be related to #123.

rudy-6-4:
triple quote """ or ''' ending a multiline string are not checked
correctly if line ends with comments or spacing or anything


rudy-6-4:
A better fix is coming with tests also.

PR #309, opened 2020-08-07
Fix #308: Add __deepcopy__ method to the `tz.TomlTz` class

### LINKED ISSUE #308: Dicts from parsed datetime with a timezone cannot be deep-copied
The resulting dict from parsing a datetime with a timezone cannot be deep-copied using
the stdlib function `copy.deepcopy()`.

## Example

```python
import copy
import toml

o = toml.loads("dob = 1979-05-27T07:32:00-08:00")
copy.deepcopy(o)
```

## What happens now

```python
TypeError: __init__() missing 1 required positional argument: 'toml_offset'
```

## What should happen

The dict is copied and the timezone of the copy has the same offset but it is a different object.

jpsca:
Without this method, the dict that results from parsing a datetime with a timezone cannot be deep-copied using the stdlib `copy.deepcopy()` function.

Example:

```python
import copy
import toml

o = toml.loads("dob = 1979-05-27T07:32:00-08:00")
# {'dob': datetime.datetime(1979, 5, 27, 7, 32, tzinfo=<toml.tz.TomlTz object at 0x7f4782efb3d0>)}
copy.deepcopy(o)
```

This currently raises a `TypeError`.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -271,3 +271,12 @@ b-comment = "a is 3"
                    encoder=toml.TomlPreserveCommentEncoder())
 
     assert len(s) == len(test_str) and sorted(test_str) == sorted(s)
+
+
+def test_deepcopy_timezone():
+    import copy
+
+    o = toml.loads("dob = 1979-05-24T07:32:00-08:00")
+    o2 = copy.deepcopy(o)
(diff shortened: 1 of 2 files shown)


uiri:
Thank you for your pull request! Merged.

PR #315, opened 2020-09-08
show meaningful error message for wrong case booleans

abred:
according to the spec booleans are all lowercase, when a capitalized boolean is used, the error message is not very informative ("This float doesn't have a leading digit")


uiri:
Thank you for the pull request! Merged.

PR #318, opened 2020-09-13
setup.py: install type stubs

FFY00:
Signed-off-by: Filipe Laíns <lains@archlinux.org>


setup.py
```diff
@@ -16,6 +16,7 @@
     author_email="uiri@xqz.ca",
     url="https://github.com/uiri/toml",
     packages=['toml'],
+    package_data={"toml": ["__init__.pyi"]},
```
dvzrv, line 19: I would make this more generic by doing a `"*.pyi"` as it is not out of the question, that more files will be created in the future.

Issue #285, opened 2020-02-14
Dump fails with UserDict


acetylen:
* Python version: 3.7.3
* toml version: 0.10.0

It seems like `UserDict` and subclasses thereof are interpreted as objects and skipped by the dumper when they occur below the top-level of an input dictionary.

test.toml:
```toml
global_key = "test"

[category]
name = "foo"
index = 75

[category.thing]
item = "thing"
value = 36.3

[category.device]
item = "device"
keys = 8
```

Test:
```python
from collections import UserDict

import toml

a = toml.load("test.toml")
b = toml.load("test.toml", _dict=UserDict)

ad = toml.dumps(a)
bd = toml.dumps(b)

print(f'{type(a).__name__}:\n"""\n{ad}"""')
print("-" * 50)
print(f'{type(b).__name__}:\n"""\n{bd}"""')
```
Output:
```
dict:
"""
global_key = "test"

[category]
name = "foo"
index = 75

[category.thing]
item = "thing"
value = 36.3

[category.device]
item = "device"
keys = 8
"""
--------------------------------------------------
UserDict:
"""
global_key = "test"
category = [ "name", "index", "thing", "device",]
"""
```

PR #335, opened 2020-11-05
Issue #285

bradermacher:
in dump_inline_table change the isinstance check to look for collections.abc.Mapping, not for dict. 

This enables UserDict and other classes that derive from collections.abc.Mapping, in addition to dict.


bradermacher:
superceded

PR #347, opened 2021-01-06
fix: py.typed

henryiii:
This package is missing a py.typed file, so downstream projects will not
show up in mypy or other type checkers, even though it has nice .pyi
files. This adds a the token file, which will cause mypy to pick it up
and use it. See [PEP 0561](https://www.python.org/dev/peps/pep-0561/).

It was also missing the .pyi files from the wheel (noticeable when running
`pip install build && python -m build`), added `include_package_data` to
fix that.

Edit: Actually I ran `pipx install build && pyproject-build`, but that's the same thing.

Closes #318

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,7 @@
 include LICENSE
 include README.rst
 include toml/*pyi
+include toml/py.typed
 include tox.ini
 include test.toml
 recursive-include tests *.py *.sh
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,3 +3,7 @@ universal = True
 
 [metadata]
 license_file = LICENSE
+
+[options]
+zip_safe = False
+include_package_data = True


henryiii:
This was attempted before in #318, apparently, but the py.typed fix as missing from that, which is quite important.

kevinjacobs-progenity:
Some of the type stubs look wrong too.  e.g. 
```
def load(f: Union[str, list, IO[str]],
         _dict: Type[MutableMapping[str, Any]] = ...,
         decoder: TomlDecoder = ...) \
         -> MutableMapping[str, Any]: ...
```
should be
```
def load(
    f: Union[str, list, IO[str]],
   _dict: Optional[Type[MutableMapping[str, Any]]] = None,
    decoder: Optional[TomlDecoder] = None,
) -> MutableMapping[str, Any]: ...
```
I'm happy to work on this.  Any objects to pushing my updates to this PR or should I open a new one?

henryiii:
I'm perfectly happy to have pushes (sorry I missed this before!). One of the reasons this may have worked in the past is that type shed had `toml` in it, but that's going away in the future (0.900), so having these would be especially useful soon!

wimax-grapl:
I too have come across this, would be really nice to get it in (and get out of Typeshed!)

henryiii:
Tomli and tomllib (in Python 3.11) have replaces this. So closing.

PR #365, opened 2021-05-26
Remove check for homogeneous array

sethcall:
In conformance with latest TOML spec...

### DIFF
--- a/toml/decoder.py
+++ b/toml/decoder.py
@@ -1024,10 +1024,7 @@ class TomlDecoder(object):
             a[i] = a[i].strip()
             if a[i] != '':
                 nval, ntype = self.load_value(a[i])
-                if atype:
-                    if ntype != atype:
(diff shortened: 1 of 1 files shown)


ksunden:
My reading of the code indicates that with this change `atype` is not at all _used_ anywhere and so the variable can simply be removed entirely (and then also `ntype` for that matter (i.e. doing `nval, _ = self.load_value(a[i])` to indicate that the returned value is not used))

zoj613:
ping @uiri

sudo-nano:
@uiri Still running into this issue in 2025. Please merge the PR.

sudo-nano:
It appears that a commit fixing this issue was introduced here 6d51c0ac52d55da4d0509a4006648d66cacab454 in 2023, but has not been pushed to release in the two years since. It seems like this library is unmaintained, and people should perhaps switch to `tomli` or another library.

Issue #368, opened 2021-06-25
Lone quote in string isn't parsed correctly


Aplet123:
Something like the following:
```
a = "\""
```
Should parse as `{"a": "\""}` but instead parses as `{"a": ""}`. Minimal example:
```py
toml.loads(toml.dumps({"a": "\""}))
# {"a": ""}
```

Holzhaus:
I can reproduce this.

PR #373, opened 2021-07-13
Fix #369: Lone quote in string isn't parsed correctly

arturoeanton:
Fix https://github.com/uiri/toml/issues/368  
Add test 
add vscode folder to .gitignore


.gitignore
```diff
@@ -112,3 +112,6 @@ venv.bak/
 auto-save-list
 tramp
 .\#*
+
+# VScode
+.vscode
```
Holzhaus, line 117: IMHO this doesn't belong into this PR. I suggest to use `git rebase` to remove the last two commits.

PR #375, opened 2021-07-18
Comments that Precede TOML lines 

amunchet:
I have a rather large TOML file template that I did not generate (Telegraf's configuration file).  The comments for each section and property precede the relevant bits - this PR allows a flag to be set in the decoder to properly retain these comments.  Currently `toml` looks for comments that immediately follow - the default behavior is not changed and the relevant tests currently pass.  

I hope this helps someone else trying to generate documentation from TOML comments.


wimax-grapl:
(Thanks for doing this PR, I'd love to have this as well!)

PR #410, opened 2023-01-17
Support pickle for TomlTz. Fix #407

### LINKED ISSUE #407: Loaded datetimes with timezone offset cannot be pickled correctly
I've tried pickling an object loaded from a toml file that contains a datetime with a timezone offset.

Here is a minimal example:
```python
>>> import toml, pickle
>>> toml_content = """
dt_with_tz = 2001-01-01 00:00:00+01:00
"""
>>> loaded_from_toml = toml.loads(toml_content)
>>> pickled = pickle.dumps(loaded_from_toml)
>>> print(pickle.loads(pickled))
  File "<stdin>", line 1, in <module>
TypeError: TomlTz.__init__() missing 1 required positional argument: 'toml_offset'
```

I have not looked deeper into the pickle data or how pickle accesses sub-objects. And yes, I know that this is a great combination, TOML+pickle.

KosukeMizuno:
Related to #407

In previous version, combining toml and pickle do not work. We can pickle them, but it cannot be loaded becauseTomlTz objects were not able to be pickled.
I made TomlTz picklable.

Procedure to reproduce the previous problem:
1. prepare a toml file including a datetime value with timezone
2. load the toml file by `toml.load()`
3. (do some editing)
4. update the toml file by `toml.dump()` and save as pickle
5. the toml file can be opened, but the pickled file could not be opened.

### DIFF
--- a/toml/tz.py
+++ b/toml/tz.py
@@ -11,6 +11,9 @@ class TomlTz(tzinfo):
         self._hours = int(self._raw_offset[1:3])
         self._minutes = int(self._raw_offset[4:6])
 
+    def __getinitargs__(self):
+        return (self._raw_offset,)
+
     def __deepcopy__(self, memo):
         return self.__class__(self._raw_offset)
 


uiri:
Awesome fix. Thanks!

PR #427, opened 2023-08-26
Iterate over all files in toml-test/tests, including sub dirs

hliang:
Tests in toml-test have been moved into subdirectories (https://github.com/BurntSushi/toml-test/pull/80). Current unit tests ([test_valid_tests](https://github.com/uiri/toml/blob/5706d3155f4da8f3b84875f80bfe0dfc6565f61f/tests/test_api.py#L47), [test_invalid_tests](https://github.com/uiri/toml/blob/5706d3155f4da8f3b84875f80bfe0dfc6565f61f/tests/test_api.py#L163)) are unable to pick up toml files in subdirectories. 

This PR change will allow the unit tests to traverse the whole directory tree (`toml-test/tests/valid/` and `toml-test/tests/invalid/`), including sub directories.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -46,11 +46,13 @@ def test_bug_196():
 
 def test_valid_tests():
     valid_dir = "toml-test/tests/valid/"
-    for f in os.listdir(valid_dir):
-        if not f.endswith("toml"):
-            continue
-        with open(os.path.join(valid_dir, f)) as fh:
-            toml.dumps(toml.load(fh))
+    for path, sub_dirs, files in os.walk(valid_dir):
+        for name in files:
+            f = os.path.join(path, name)
+            if not f.endswith("toml"):
+                continue
+            with open(f) as fh:
+                toml.dumps(toml.load(fh))
 
 
 def test_circular_ref():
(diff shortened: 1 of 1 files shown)


Warchant (approved):

Warchant:
@uiri

arp242:
It should read the `tests/files-toml-1.0.0` file and run only those tests, otherwise it will also run TOML 1.1 test files.

Or alternatively use the `toml-test` binary, which seems fast enough:

```
% time (toml-test ./tests/decoding_test.py | tail -n3)
toml-test [./tests/decoding_test.py]: using embedded tests
  valid tests: 132 passed, 25 failed
invalid tests: 230 passed, 53 failed
( toml-test ./tests/decoding_test.py | tail -n3; )  22.02s user 3.80s system 769% cpu 3.354 total
```

uiri:
Is there a list for TOML 0.5.0 tests until we get 1.0 support :) ?

I think it is fine for now for it to run all the tests.

arp242:
> Is there a list for TOML 0.5.0 tests until we get 1.0 support :) ?

Unfortunately when I updated toml-test to support TOML 1.0 a few years ago I didn't keep track of which tests are for TOML 1.0 and which are for <=0.5. I don't mind adding this as such, but figuring out which test is for which version is some work, and it's not something that's come up that often.

But if you have a list of tests that are for TOML 0.5 only then I'll be happy to add a `toml 0.5` flag. It's not a lot of work except for figuring out which tests.

> I think it is fine for now for it to run all the tests.

The problem is some tests are directly contradictory; for example for TOML 1.0 it will test that newlines and trailing commas re forbidden, as they're explicitly disallowed. But for TOML 1.1 it will test they're allowed, as the restrictions surrounding that are not relaxed.

PR #431, opened 2023-10-08
Fix #430: list index out of range

### LINKED ISSUE #430: List index out of range + unparseable UTF8 chars
After `toml` was not able to parse `no-break-space` (`u'\uc2a0'`), I became curious - what other utf8 chars cannot be parsed by `toml`.

Came up with a following script:
```python
import struct
import sys
import traceback

import toml


def find_all_chars_that_are_unsupported_by_toml_parser():
    bad = []
    for a in range(0, 256):
        for b in range(0, 256):
            c = None
            try:
                c = struct.pack("BB", a, b)
                toml.loads(f"{c.decode('utf-8')}")
            except UnicodeDecodeError:
                pass
            except IndexError as e:
                traceback.print_exc()
                print(c)
                sys.exit(1)
            except toml.TomlDecodeError as e:
                print(e)
                bad.append(c)

    return bad



bad = find_all_chars_that_are_unsupported_by_toml_parser()
print(len(bad))
```

One interesting thing is that when `c=b'\x00\r'`, then toml.loads fails with IndexError:
```
    toml.loads(f"{c.decode('utf-8')}")
  File "C:\W\cli\.venv\lib\site-packages\toml\decoder.py", line 207, in loads
    if item == '\r' and sl[i + 1] == '\n':
IndexError: list index out of range
```

This issue is a request to fix IndexError. 
If you replace this:
```diff
-toml.loads(f"{c.decode('utf-8')}")
+toml.loads(f"c = '{c.decode('utf-8')}'")
```
Then IndexError is gone, but there are 506 (on Windows) symbols cannot be parsed by `toml`.

Warchant:
+ add parenthesis in encoder.py, because without them py3.10 did not want to run the code:
```
Generator expressions must be parenthesized if not sole argument
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ nosetests.xml
 coverage.xml
 *.cover
 .hypothesis/
+.pytest_cache
 
 # Translations
 *.mo
@@ -112,3 +113,6 @@ venv.bak/
 auto-save-list
 tramp
 .\#*
+
+# vscode
+.vscode
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -17,6 +17,12 @@ c = 2
 TEST_DICT = {"a": {"b": 1, "c": 2}}
 
 
+def test_bug_430():
+    # https://github.com/uiri/toml/issues/430 - IndexError
+    with pytest.raises(toml.TomlDecodeError, match="Key name found without value."):
+        toml.loads('\x00\r')
+
+
 def test_bug_148():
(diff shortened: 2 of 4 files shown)


uiri:
Thank you for finding the bug in #430 and for the fix!

Warchant:
@uiri would you be interested in migration from travis to github actions?

uiri:
@Warchant Yes, I would be interested in a migration from Travis to anything that is free for open source (and where this project would qualify as open source). I initially chose Travis on that basis, and then they changed their pricing model at some point over the last decade.

PR #432, opened 2023-10-09
Add github CI

Warchant:
Run: https://github.com/Warchant/toml/actions/runs/6459923999


Warchant:
And @uiri

uiri:
Thanks! I think the golang toml-test suite is more important than the tox tests, but this library hasn't implemented all of TOML 1.0 yet, so it is likely that it doesn't pass the full toml-test suite.

PR #433, opened 2023-10-10
Fix toml-test scripts

arp242:
Update tests/decoding_test.py to use the format used by the current version of toml-test. I also removed Python 2 support (my distro doesn't ship it, so I can't test it).

Also add tests/encoding_test.py for encoder tests

Still quite a few failures, mostly because it looks like it hasn't been updated to use 1.0 yet:

    % toml-test ./decoding_test.py | tail -n2
      valid tests: 139 passed, 18 failed
    invalid tests: 230 passed, 53 failed

    % toml-test ./encoding_test.py -encoder | tail -n1
    encoder tests: 146 passed, 11 failed


.github/workflows/ci.yaml
```diff
@@ -22,20 +22,20 @@ jobs:
...
       # - name: Run go tests
       #   run: |
-      #       pip -q install setuptools
-      #       python setup.py install
-      #       ./toml-test/cmd/toml-test/toml-test -i ./tests -g ./tests/gotests.sh
+      #       export PYTHONPATH=.
+      #       e=0
+      #       toml-test          ./tests/decoding_test.py || e=1
+      #       toml-test -encoder ./tests/encoding_test.py || e=1
+      #       exit $e
```
arp242, line 41: Didn't test this bit, but should be alright. You can add expected test failures with something like:

    skip_dec=(
        -skip test1
        -skip test2
    )
    toml-test -skip $skip_dec[@] ./tests/decoding_test.py || e=1

Generate a list with something like:

    toml-test -color never ./tests/decoding_test.py |& grep -a FAIL | sed 's/FAIL/-skip/'

tests/decoding_test.py
```diff
@@ -1,56 +1,37 @@
...
-    elif isinstance(value, basestring):
+        return [tag(v) for v in value]
+    elif isinstance(value, str):
         return {'type': 'string', 'value': value}
     elif isinstance(value, bool):
         return {'type': 'bool', 'value': str(value).lower()}
     elif isinstance(value, int):
         return {'type': 'integer', 'value': str(value)}
-    elif isinstance(value, long):
+    elif isinstance(value, int):
```
uiri, line 22: Looks like this is a duplicate elif block with python2 long changed to int for python3.

arp242: Cheers, fixed it.

uiri:
Merged. Thanks!

PR #434, opened 2023-10-10
Allow using heterogeneous arrays

arp242:
This is valid since TOML 1.0


uiri:
Thanks!
This is (was?) a very controversial change, I am sure ;)

PR #435, opened 2023-10-10
Treat "Z" offset in datetime as case-insensitive

arp242:
This should be case-insensitive; e.g. this test failed:

	% toml-test ./tests/decoding_test.py -run valid/datetime/datetime
	FAIL valid/datetime/datetime
	     Key "lower" is not "datetime" but "datetime-local":
	       Expected:     map[string]any{"type":"datetime", "value":"1987-07-05T17:45:00Z"}
	       Your encoder: map[string]any{"type":"datetime-local", "value":"1987-07-05T17:45:00"}


uiri:
Thank you for this fix!

